PR #7, opened 2018-05-03
Allow accessing lane counts from Run Scanner

apmasell:
This intentionally ignores the joined lane problem until we have a more systematic solution


source-pinery/src/main/java/ca/on/oicr/gsi/shesmu/variables/provenance/FileProvenanceVariablesSource.java
```diff
@@ -140,6 +140,7 @@ public FileProvenanceVariablesSource() throws IOException {
 				final AtomicInteger badVersions = new AtomicInteger();
 				final Map<String, Integer> badSetCounts = new TreeMap<>();
 				cache = Utils.stream(client.getFileProvenance(PROVENANCE_FILTER))//
+						.filter(fp -> fp.getSkip() == null || fp.getSkip().equals("false"))//
```
callunity: Do you want to add the boolean `false` in here so you don't have to update it later?

apmasell: It won't compile if that's the case. It's also going to change a lot of pipedev code, so I'll wait to see how that shakes out.

PR #9, opened 2018-05-03
Separate provenance

apmasell:


source-pinery/src/main/java/ca/on/oicr/gsi/shesmu/variables/provenance/PineryProvenanceVariablesSource.java
```diff
@@ -0,0 +1,211 @@
...
+		}
+	}
+
+	private static final Gauge badSetMap = Gauge
+			.build("shesmu_run_provenance_bad_set",
+					"The number of provenace records with sets not containing exactly one item.")
+			.labelNames("target", "property", "reason").register();
+
+	private static final Gauge count = Gauge
+			.build("shesmu_run_provenance_last_count", "The number of items from Provenance occured.").register();
```
callunity: "occured"? (I'm not quite understanding the human-readable explanation)

apmasell:
Okay. I have this all muddled in my head: I'm just going to call this Pinery provenance and I've updated all the Prometheus variables to match. I'll do the same for the commit name when I merge.

PR #17, opened 2018-05-15
Attach descriptions to functions and constants for debugging

apmasell:


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/ExpressionNodeFunctionCall.java
```diff
@@ -16,6 +16,11 @@
...
 	private static final FunctionDefinition BROKEN_FUCNTION = new FunctionDefinition() {
 
+		@Override
+		public String description() {
+			return "Undefined function";
+		}
+
 		@Override
 		public String name() {
 			return "💔";
```
callunity: :laughing:

PR #26, opened 2018-06-08
Add and update Swagger documents and licence

apmasell:


COPYING
```diff
@@ -0,0 +1,19 @@
+Copyright 2018 The Ontario Institute for Cancer Research
```
callunity: I think we usually do our copyright as `Genome Sequence Informatics`, but Morgan would be the person to confirm that with.

callunity: This looks like the MIT license to me; is there a particular reason to not call it that?

PR #44, opened 2018-07-05
Check stream consumption when subsampling

apmasell:


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/SampleNodeFixedWithCondition.java
```diff
@@ -75,6 +75,16 @@ public void collectFreeVariables(Set<String> names) {
...
 	}
 
+	@Override
+	public Consumption consumptionCheck(Consumption previous, Consumer<String> errorHandler) {
+		if (previous == Consumption.VARIABLE) {
+			errorHandler.accept(String.format("%d:%d: Cannot consume a fixed number of items after a variable number.",
+					limitExpression.line(), limitExpression.column()));
+			return Consumption.BAD;
+		}
+		return previous;
```
callunity: Why does this `consumptionCheck` (and the one in SampleNodeFixed) return `previous`, while the one in SampleNodeSquish returns `Consumption.VARIABLE`? (aka why wouldn't the Fixed ones return `Consumption.FIXED`?)

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/SampleNodeSquish.java
```diff
@@ -31,6 +31,14 @@ public void collectFreeVariables(Set<String> names) {
 
 	}
 
+	@Override
+	public Consumption consumptionCheck(Consumption previous, Consumer<String> errorHandler) {
+		if (previous == Consumption.BAD) {
+			return Consumption.BAD;
+		}
+		return Consumption.VARIABLE;
```
evelii: so does this mean that it can guarantee that we only allow `Squish` to be called after `Fixed` or `FixedWithCondition`?

evelii: (and `Squish` can be the first subsampler without being followed by `Fixed` or `FixedWithCondition`.)

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/SampleNodeFixedWithCondition.java
callunity: I'm in favour of explicit switch blocks.

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/SampleNodeSquish.java
callunity: Yes, that's correct.

PR #51, opened 2018-07-12
Add emergency stop button that engages a throttler

apmasell:


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
```diff
@@ -330,6 +351,8 @@ public Server(int port) throws IOException {
 			}
 		});
 
+		add("/resume", new EmergencyThrottlerHandler(false));
+		add("/stopstopstop", new EmergencyThrottlerHandler(true));
```
morgantaschuk: :100:

PR #52, opened 2018-07-13
Include type information when setting up a dumper

apmasell:


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/OliveNode.java
```diff
@@ -139,12 +140,15 @@ public abstract boolean resolve(InputFormatDefinition inputFormatDefinition, Con
 	 *
 	 * This does the clauses and
 	 * {@link #resolveDefinitionsExtra(Map, Function, Function, Consumer)}
+	 * 
+	 * @param dumpers
+	 *            TODO
```
callunity: Stray TODO?

apmasell: Shakes fist at Eclipse.

PR #57, opened 2018-07-19
Make `List` operation consistent with others

apmasell:
Every other `For` operation has an expression except `Count`.


language.md
```diff
@@ -521,9 +521,9 @@ list separated by the value of _delimexpr_, which must also be a string.
   joining.
 
 #### List
-- `List`
+- `List` _expr_
 
-Collects all the items into a list.
+Evaluates _expr_ for every item and collects all the unique into a list.
```
callunity: Why only the unique values?

apmasell: Because everything is really a set.

PR #64, opened 2018-07-24
Switch to AutoUpdatingDirectory for PrometheusThrottler

apmasell:


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/throttler/prometheus/PrometheusThrottler.java
```diff
@@ -32,57 +31,76 @@
...
+							.anyMatch(alert -> alert.matches(configuration.getEnvironment(), services))))
+					.orElse(false);
+		}
 
-	private static final Gauge ok = Gauge
-			.build("shesmu_prom_throttler_status", "Whether the last call to Alert Manager was successful.").register();
+		@Override
+		protected Optional<Integer> update(Configuration value) {
+			configuration = Optional.of(value);
+			return Optional.empty();
```
callunity: Why does it return `Optional.empty()`?

apmasell: That's if you succeeded. Otherwise, the number of minutes to try again.

PR #80, opened 2018-08-01
Add titles to histogram in action dashboard

apmasell:


shesmu-server/src/main/resources/shesmu.js
```diff
@@ -559,6 +559,17 @@ function propertyFilterMaker(name) {
   }
 }
 
+function nameForBin(name) {
+  switch (name) {
+    case "added":
+      return "Last Generated by an Olive";
+    case "checked":
+      return "Last Checked Action Status";
```
callunity: This sounds like it should be a string (the Status) but that's not correct...

shesmu-server/src/main/resources/shesmu.js
```diff
@@ -559,6 +559,17 @@ function propertyFilterMaker(name) {
   }
 }
 
+function nameForBin(name) {
+  switch (name) {
+    case "added":
+      return "Last Generated by an Olive";
```
callunity: Time Since Action Was Last Generated by an Olive ?

shesmu-server/src/main/resources/shesmu.js
callunity: To be even more obtuse, what does it mean for an action to be "Checked"? Is it the action that is checked, or something else? There isn't anything in the documentation about this term...

apmasell: How this instead?

callunity: Could probably drop one of the two "Last"s in the second one, but yeah, much clearer.
Could you also put variations on these titles up at the top of the page, where the user fills in date ranges?

PR #103, opened 2018-08-29
Nabu

apmasell:
This depends on there being an endpoint in Nabu that dumps everything as one big array that does not yet exist in Nabu.


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/input/nabu/NabuValue.java
```diff
@@ -0,0 +1,84 @@
...
+	public boolean skip() {
+		return skip;
+	}
+
+	@Export(type = "s")
+	public String stalestatus() {
+		return stalestatus;
+	}
+
+	@Export(type = "as")
```
callunity: I believe this is actually `type = "ai"`, as it's an array of upstream file SWIDs

PR #113, opened 2018-09-17
Update Nabu input format to match new Nabu endpoint

apmasell:


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/input/nabu/NabuJsonRepository.java
```diff
@@ -21,9 +19,14 @@ public NabuJsonRepository() {
...
-		for (JsonNode value : node.get("upstream")) {
-			upstream.add(value.asLong());
+		final JsonNode qcstatus = node.get("qcstatus");
+		String status;
+		if (qcstatus == null || qcstatus.isNull()) {
+			status = "PENDING";
+		} else if (qcstatus.isBoolean()) {
+			status = qcstatus.asBoolean() ? "PASSED" : "FAILED";
+		} else {
+			status = qcstatus.asText();
```
callunity: What are the chances that this'll return `true` or `false`?

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/input/nabu/NabuJsonRepository.java
```diff
@@ -38,14 +41,11 @@ protected NabuValue convert(ObjectNode node) {
 		return new NabuValue(//
 				node.get("fileswid").asInt(), //
 				node.get("filepath").asText(), //
-				node.get("qcstatus").asText(), //
+				status, //
 				node.get("username").asText(), //
-				node.get("comment").asText(), //
+				node.get("comment").asText(""), //
```
callunity: Why the empty double quotes?

apmasell: That provides a default if missing or null and Shesmu can't do null, so empty string is close enough.

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/input/nabu/NabuJsonRepository.java
apmasell: I hope not. The reason for this (and the time handling above) is that the Shesmu `/input/nabu` endpoint is going to output the data in Shesmu's format, where this is a string, and not a Boolean. So, this handles the case when it's coming in that pre-processed format.

PR #116, opened 2018-09-19
Update vim syntax highlighting

callunity:


vim/syntax.vim
```diff
@@ -66,6 +67,9 @@ syn keyword shesmuKeyword Where
 syn keyword shesmuKeyword While
 syn keyword shesmuKeyword With
 syn keyword shesmuBool False True
+syn keyword shesmuKeyword json_signature
```
apmasell: These aren't guaranteed to be there, so I'm not sure they're a good idea to have in there, but please change to `shesmuConstant`.

PR #118, opened 2018-09-21
Change Matches to Call and add TypeAlias

apmasell:


vim/syntax.vim
```diff
@@ -19,6 +19,7 @@ syn keyword shesmuKeyword All
 syn keyword shesmuKeyword Annotations
 syn keyword shesmuKeyword Any
 syn keyword shesmuKeyword By
+syn keyword shesmuKeyword Call
```
callunity: Can remove `Call`

callunity:
LGTM

PR #128, opened 2018-10-03
Add package info and fix tests

apmasell:


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/package-info.java
```diff
@@ -0,0 +1,7 @@
+/**
+ * Shesmu: a decision action server
+ * 
+ * This package contains the primary interfaces and base classes use for
+ * extending Shesmu to interface with other system
```
callunity: `system` -> `systems`

PR #132, opened 2018-10-11
Use common status page package

apmasell:


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/CompiledGenerator.java
```diff
@@ -115,11 +127,8 @@ public CompiledGenerator(Supplier<Stream<FunctionDefinition>> functions, Supplie
 	 *
 	 * @return
 	 */
-	public String errorHtml() {
-		return scripts()//
-				.filter(script -> script.errors.length() > 0)//
-				.map(script -> String.format("<p><b>%s</b></p><p>%s</p>", script.fileName.toString(), script.errors))//
-				.collect(Collectors.joining());
+	public void errorHtml(SectionRenderer renderer) {
+		scripts().forEach(script -> script.errorHtml(renderer));
```
callunity: Is the formatting of these errors set in `server-utils`?

apmasell: Yes. I've changed the formatting a bit. They appear as rows in the first table.

PR #133, opened 2018-10-15
Fix spelling of statuses

apmasell:


plugin/jira/src/main/java/ca/on/oicr/gsi/shesmu/jira/Configuration.java
```diff
@@ -43,8 +43,8 @@ public void setCloseActions(List<String> closeActions) {
 		this.closeActions = closeActions;
 	}
 
-	public void setClosedStatues(List<String> closedStatues) {
-		this.closedStatues = closedStatues;
+	public void setClosedStatsues(List<String> closedStatuses) {
```
callunity: `setClosedStatsues` -> `setClosedStatuses`

PR #134, opened 2018-10-18
Add a separate format for Pinery lanes

apmasell:
Rather than making fake analysis provenance

### DIFF
--- a/plugin/niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/RemotePineryIUSRepository.java
+++ b/plugin/niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/RemotePineryIUSRepository.java
@@ -1,0 +1,244 @@
+package ca.on.oicr.gsi.shesmu.pinery;
+
+import java.nio.file.Path;
+import java.time.Duration;
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Collections;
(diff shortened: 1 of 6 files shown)


plugin/niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/RemotePineryIUSRepository.java
```diff
@@ -0,0 +1,244 @@
...
+		}
+	}
+
+	private static final Gauge badSetMap = Gauge
+			.build("shesmu_pinery_ius_bad_set",
+					"The number of provenace records with sets not containing exactly one item.")
+			.labelNames("target", "property", "reason").register();
+
+	private static final Gauge count = Gauge
+			.build("shesmu_pinery_ius_last_count", "The number of lanes and samples from Pinery.").labelNames("target")
```
callunity: Is this intended to count the sum total of lanes + samples? Or samples-in-lanes (IUSes)?

callunity: If it's not counting IUSes, can the label be changed?

apmasell: Sure. Just `pinery`. It's cleaner that way.

PR #137, opened 2018-10-19
Change `PartitionCount`'s return type from tuple to object

apmasell:
Rather than return two counts in a tuple, return two counts with helpful
labels.

### DIFF
--- a/language.md
+++ b/language.md
@@ -118,8 +118,9 @@ In total, the collectors in a `Group` operation are:
 - `Max` and `Min` to collect the most extreme value; if none are collected, the
   group is rejected
 - `Count` to count the number of matched rows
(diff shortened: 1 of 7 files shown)


language.md
```diff
@@ -721,8 +722,9 @@ condition specified in _expr_, which must return a Boolean.
 #### Partitioned Counter
 - `PartitionCount` _expr_
 
-Produces a tuple with two elements: the first is the number of items for which
-_expr_ was true, the second is the number of items for which _expr_ was false.
+Produces an with two field: `matched_count` is the number of items for which
```
morgantaschuk: an what?

apmasell: Fixed.

PR #144, opened 2018-10-24
Improve type parser in UI

apmasell:
* This can parse types and signatures
* It displays both the type and the signature
* It provides the input-format and signature-format types

### DIFF
--- a/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/util/server/TypeParseResponse.java
+++ b/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/util/server/TypeParseResponse.java
@@ -1,0 +1,29 @@
+package ca.on.oicr.gsi.shesmu.util.server;
+
+import ca.on.oicr.gsi.shesmu.Imyhat;
+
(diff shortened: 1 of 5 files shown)


apmasell:
Added.

apmasell:
Okay. I've improved the explanation for this page and made it easier to use. As part of a longer term changing, I'm renaming “type signature” to “type descriptor”. This changes how it appears in the type page, but not in the rest of the code. That will happen in another PR.

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/util/server/TypeParseResponse.java
```diff
@@ -0,0 +1,29 @@
...
+	public TypeParseResponse(Imyhat input) {
+		human = input.name();
+		descriptor = input.signature();
+	}
+
+	public String getDescriptor() {
+		return descriptor;
+	}
+
+	public String getHuman() {
```
djcooke: 'name' or 'humanName' might be clearer :alien:

PR #145, opened 2018-10-25
Use RuntimeBinding for actions too

apmasell:


plugin/guanyin/src/main/java/ca/on/oicr/gsi/shesmu/guanyin/ReportActionRepository.java
```diff
@@ -76,14 +90,22 @@ public void emit(SectionRenderer renderer) throws XMLStreamException {
 			}
 		}
 
+		public String 观音Url() {
```
iainrb: Is Guanyin supposed to be rendered in Chinese here?

PR #146, opened 2018-10-25
Simplify exporting Shesmu variables via annotations

apmasell:


plugin/nabu/src/main/java/ca/on/oicr/gsi/shesmu/nabu/NabuValue.java
```diff
@@ -25,37 +25,37 @@ public NabuValue(long fileswid, String filepath, String qcstatus, String usernam
...
 	}
 
-	@Export(type = "s")
+	@Export
 	public String qcstatus() {
 		return qcstatus;
 	}
 
-	@Export(type = "s")
+	@Export
```
callunity: Should these be changed to `@ShesmuVariable` like in the other files?

plugin/niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PineryIUSValue.java
```diff
@@ -74,12 +74,12 @@ public PineryIUSValue(//
 		this.completed_date = completed_date;
 	}
 
-	@Export(type = "d")
+	@ShesmuVariable(type = "d")
```
callunity: Do `type = "s"` and `type = "d"` still need to be explicitly declared? (They're also declared in `InnerTestValue` and `TestValue` below)
Could it please be Javadoc'd somewhere which types need to be explicitly declared?

plugin/nabu/src/main/java/ca/on/oicr/gsi/shesmu/nabu/NabuValue.java
apmasell: Yes...I'm not sure why they didn't.

plugin/niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PineryIUSValue.java
apmasell: No, but it does a check that if you specify the type, it matches the actual return type, so I'm leaving thsese in for testing purposes.

callunity: I can see the value in leaving them in `InnerTestValue` and `TestValue` for testing purposes, but I'm concerned that leaving them in `PineryIUSValue` makes for potentially confusing inconsistent styles between the plugin code files.

apmasell: Ooops. Didn't remember that one. Fixed.

implementation.md
```diff
@@ -200,6 +200,10 @@ the `Input` instruction. This class must be annotated with
 1. Create new classes that provide data which implement _R_ and are annotated
 with `@MetaInfServices(`_R_`)`.
 
+For each variable, Shesmu can try to infer the type from the return type of the
+method. If this is not possible, it must be specified in the `type` property of
```
djcooke: Type can be inferred for boolean, long, Instant, and String, but Set and Tuple types must be specified, right? Please note that here

apmasell: Fixed. Thanks.

PR #148, opened 2018-10-29
Simplify exporting Shesmu actions, constants, and functions via annotations

apmasell:


implementation.md
```diff
@@ -353,13 +403,13 @@ To create an action:
 1. Annotate this class with `@MetaInfServices`.
```
morgantaschuk: The _type_ thing above confused me. I was expecting an annotation or something. Maybe provide an example

shesmu-server/src/test/resources/compiler/bad-missing-action.errors
```diff
@@ -1 +1 @@
-3:6: Unknown action for “nothing”.
+3:6: Unknown action for “notathing”.
```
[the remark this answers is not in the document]
apmasell: Yes, but changed now for more clarity.

implementation.md
```diff
@@ -217,41 +215,88 @@ instance, storing it as a JSON file, then using that on another instance.
...
-something more complicated is desired, extend the `Constant` class and write
-arbitrary bytecode in the `load` method. The code must not change any value on
-the stack and must leave exactly one value of the correct type on the stack.
-
-The `RuntimeBinding` class can be used to create constants as method calls on
-the instance of a class.
-
-### Functions
+##### Manual Implementation
+When directly implementing the `DefinitionRepository` interface, create
```
morgantaschuk: Perhaps indicate why you would need to implement manually. What use case do you see?

implementation.md
```diff
@@ -217,41 +215,88 @@ instance, storing it as a JSON file, then using that on another instance.
...
-To provide functions:
-
-1. Create a class that implements `FunctionRepository`.
-1. Annotate this class with `@MetaInfServices`.
-1. Include a no-arguments constructor for the `FunctionRepository` class.
-1. Return a stream of `FunctionDefintion` objects for the functions to be created.
-
-There are a few ways methods to generate a `FunctionDefinition`:
+##### Manual Implementation
+When directly implementing the `DefinitionRepository` interface, create
```
morgantaschuk: Same as above. Indicate why we would need to manually implement. With this section and the other manual implementation sections, I would suggest moving them after the File-Backed* sections: that way people can see if there is non-manual implementation that matches their use case before moving on to manual.

implementation.md
```diff
@@ -344,22 +402,24 @@ When queried by the user, an action can return a JSON representation. This is
 arbitrary and entirely for the benefit of users. Any information can be
 included in an appropriate way.
 
+##### Manual Implementation
+
```
morgantaschuk: As before: tell me why I would need manual + move this to the end of this section.

implementation.md
```diff
@@ -217,41 +215,127 @@ instance, storing it as a JSON file, then using that on another instance.
...
+When directly implementing the `DefinitionRepository` interface, create
+instances of `FunctionDefinition`.  There are a few ways methods to generate a
+`FunctionDefinition`:
 
 1. Create a class that implements `FunctionDefinition` and write arbitrary
-bytecode.
+bytecode. For example [`str_len` in `StandardDefinitions`](shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/core/StandardDefinitions.java).
 1. Use `FunctionDefinition.staticMethod` to create a binding for a public
 static method in a class.
 1. Use `RuntimeBinding` to create a binding for a method in an instance of a
```
iainrb: Is it worth having an explanation of what RunTimeBinding is? I notice there was one, in the deleted section headed Run Time Binding. Is that section now out of date? If not, is it a good idea to reinstate it?

apmasell: I think I should delete this advice. When each plugin was doing its own thing, they were using this class directly. Now it's basically a utility of the `FileBackMatchedDefinitionRepository`, so it shouldn't be mentioned.

iainrb: OK, makes sense.

implementation.md
```diff
@@ -344,22 +407,24 @@ When queried by the user, an action can return a JSON representation. This is
...
 1. Override all the methods for the desired behaviour as described above.
-1. Create a class _AR_ that implements `ActionRepository`.
-1. Annotate this class with `@MetaInfServices`.
-1. Include a no-arguments constructor for the `ActionRepository` class.
+1. Create a class _AR_ that implements `DefinitionRepository`.
+1. Annotate this class with `@MetaInfServices(DefinitionRepository.class)`.
+1. Include a no-arguments constructor for the `DefinitionRepository` class.
 1. Create a class _AD_ that extends `ActionDefinition`.
-1. In the constructor of _AD_, call the superconstructor with `("_json type name_`"`, Type.getType(`_A_`.class)`, Stream.of(`_parameters_`);`
+1. In the constructor of _AD_, call the superconstructor with `("`_shesmu name_`"`, Type.getType(`_A_`.class)`, Stream.of(`_parameters_`);`
```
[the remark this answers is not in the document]
apmasell: No. They were wrong in the original documentation.

implementation.md
```diff
@@ -344,22 +407,24 @@ When queried by the user, an action can return a JSON representation. This is
 arbitrary and entirely for the benefit of users. Any information can be
 included in an appropriate way.
 
+##### Manual Implementation
+
 To create an action:
 
 1. Create a class _A_ that extends `Action`.
-1. It must provide a unique _type_ name to the superconstructor that will be available for searching via the REST API.
+1. It must provide a unique _json type name_ to the superconstructor that will be available for searching via the REST API and the action dashboard. (_e.g._, `jira-open-ticket`, `nothing`, `fake`)
```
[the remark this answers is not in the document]
apmasell: I'm rewording this to be _JSON action type_ and _name for use in olives_. Is that better?

plugin/niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/gsistd/niassa/WorkflowType.java
```diff
@@ -20,47 +28,99 @@
...
 	}
 
-	public Stream<WorkflowParameterDefinition> parameters() {
+	public final Stream<ActionParameterDefinition> parameters() {
 		return Stream.of(definitions);
 	}
 
-	public Type type() {
-		return type;
+	public interface Insanity {
```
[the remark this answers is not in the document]
apmasell: I've got another PR to remove this entirely.

plugin/guanyin/src/main/java/ca/on/oicr/gsi/shesmu/guanyin/GuanyinFile.java
```diff
@@ -0,0 +1,96 @@
...
+						report.getPermittedParameters()//
+								.entrySet().stream()//
+								.map(e -> new JsonParameter(e.getKey(), Imyhat.parse(e.getValue().getType()),
+										e.getValue().isRequired())));
+			}
+			this.configuration = Optional.of(configuration);
+			return Optional.of(60);
+		} catch (final IOException e) {
+			e.printStackTrace();
+			return Optional.of(5);
```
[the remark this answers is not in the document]
djcooke: Still not sure what the return value means

PR #155, opened 2018-11-09
Move GitHub branches input repository to new caching layer

apmasell:


plugin/github/src/main/java/ca/on/oicr/gsi/shesmu/github/GitHubBranchesApiRepository.java
```diff
@@ -52,76 +86,21 @@ public void emit(SectionRenderer renderer) throws XMLStreamException {
...
 
 		@Override
 		public Optional<Integer> update(Configuration value) {
 			configuration = Optional.of(value);
+			cache.invalidate();
 			return Optional.empty();
 		}
 	}
 
-	private static final Gauge count = Gauge.build("shesmu_github_last_count", "The number of branches from GitHub.")
```
morgantaschuk: You got rid of the monitoring in this class?

apmasell: Yes. It's handled by the caching layer now.

PR #162, opened 2018-11-16
Allow creating user-defined constants

apmasell:
I had to change `Target` from a class to an interface, which generated a lot of noise.


language.md
```diff
@@ -396,8 +396,17 @@ Functions may use constants or previously defined functions (either built-in or
 defined in the file). Functions cannot be defined recursively. The return type
 is determined automatically.
 
+## Constants
+At the top level of a file, constants may be defined:
+
+    myval = 3 * 12;
+
+The body of constants may use functions defined in this file. However,
```
callunity: Maybe `A constant's definition may use functions defined in this file`? I got hung up on whether constants have bodies, but maybe that's just me.

apmasell: This was probably copy pasta from somewhere. I'll fix it.

language.md
```diff
@@ -396,8 +396,17 @@ Functions may use constants or previously defined functions (either built-in or
 defined in the file). Functions cannot be defined recursively. The return type
 is determined automatically.
 
+## Constants
+At the top level of a file, constants may be defined:
```
callunity: `At the top level of a file, ...` -> `At the top level of a file, after any TypeAliases, ...`

apmasell:
Updated.

PR #169, opened 2018-11-23
Suppress exceptions when counting input sources

apmasell:
Olive have their own exception handlers, but the counting logic did not.

### DIFF
--- a/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/CompiledGenerator.java
+++ b/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/CompiledGenerator.java
@@ -70,8 +70,9 @@ public class CompiledGenerator extends ActionGenerator {
 		@Override
 		public synchronized Optional<Integer> update() {
 			try (Timer timer = compileTime.labels(fileName.toString()).startTimer()) {
-				final HotloadingCompiler compiler = new HotloadingCompiler(SOURCES::get, DefinitionRepository::allFunctions,
-						DefinitionRepository::allActions, DefinitionRepository::allConstants);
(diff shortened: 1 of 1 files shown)


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/CompiledGenerator.java
```diff
@@ -127,9 +128,13 @@ public void errorHtml(SectionRenderer renderer) {
...
 		// Load all the input data in an attempt to cache it before any olives try to
 		// use it. This avoids making the first olive seem really slow.
-		InputFormatDefinition.formats()
-				.forEach(format -> INPUT_RECORDS.labels(format.name()).set(format.input(format.itemClass()).count()));
-		ActionGenerator.OLIVE_FLOW.clear();
+		InputFormatDefinition.formats().forEach(format -> {
+			try {
+				INPUT_RECORDS.labels(format.name()).set(format.input(format.itemClass()).count());
+			} catch (Exception e) {
+				e.printStackTrace();
```
morgantaschuk: Is it a good idea to printStackTrace here? does Shesmu have a logger library yet?

apmasell: Yes. If there's a broken input source, that error should go to the console.

There's no logger and given it runs as a standalone application, I'm not sure there's a reason to have a logger. Everything ends up in the system log if it goes to the console.

morgantaschuk: :+1:

PR #189, opened 2019-01-03
Add a guide to working on the Shesmu compiler

apmasell:


compiler-hacking.md
```diff
@@ -0,0 +1,464 @@
...
+      public static <clinit>()V
+        RETURN
+        MAXSTACK = 0
+        MAXLOCALS = 0
+    }
+
+This class, `dyn.shesmu.Program`, extends the `ActionGenerator` superclass,
+which is used by the server to execute actions. The class name is reused for
+every script, but since each is loaded by a separate class loader, they are not
+in conflict. The constructor is relatives boring; it just calls the
```
avarsava: typo

compiler-hacking.md
```diff
@@ -0,0 +1,464 @@
...
+        MAXSTACK = 0
+        MAXLOCALS = 1
+    
+      public synchronized run(Lca/on/oicr/gsi/shesmu/ActionConsumer;Ljava/util/function/Function;)V
+        ALOAD 0
+        INVOKEVIRTUAL dyn/shesmu/Program.clearGauge ()V
+
+Now there are interesting differences. First, the current time is stored for timing the olive runtime.
+
+       L0
```
avarsava: It wasn't immediately clear to me that this was an extension of the previous code sample, could the entire thing be provided beforehand or the commentary moved to code comments?

compiler-hacking.md
```diff
@@ -0,0 +1,464 @@
...
+        INVOKESTATIC java/lang/System.nanoTime ()J
+        LSTORE 3
+
+Now, a stream of the correct type must be started that will contain the right
+class. Since the supplied argument is `Function<Class<T>, Stream<T>>`, it's
+invoked with the class specified by the `Input` declaration:
```
avarsava: Why must a stream be created? 
How is the correct type determined? 
Where was that argument supplied, and how does it imply that 'it' (the stream? the argument?) should be invoked using the class from Input?

compiler-hacking.md
```diff
@@ -0,0 +1,464 @@
+# Understanding the Shemsu compiler
+This is a guide to understanding the Shesmu olive compiler. This is meant to
+explain the core components that get the language to Java bytecode.
```
avarsava: Could you detail what background knowledge (JVM, etc) is needed to understand the document?

compiler-hacking.md
```diff
@@ -0,0 +1,464 @@
...
+
+    boolean check(
+      Instant now,
+      ShesmuIntrospectionValue streamValue)
+
+By Java convention, when making a lambda, the arguments to the lambda itself
+come after the captures. In the case of Shesmu's use of stream, this always the
+_stream value_ (_i.e._, the current item in the stream).
+
+When access in the stream variable `changed`, it will generate code like:
```
avarsava: Was this supposed to have quotes around it, or is 'changed' a special term in Shesmu?

compiler-hacking.md
```diff
@@ -0,0 +1,464 @@
...
+3. Resolve variables.
+4. Check the type and stream purity of everything.
+5. Convert everything to bytecode.
+
+There are a lot of classes in the compiler, but the major ones are:
+
+- `OliveNode`, which declares each top-level element in the file (function,
+  olive, define olive, constant)
+- `OliveClauseNode`, which defines every clause in an olive
+- `ExpressionNode`, which defines all the bits an pieces of expressions
```
avarsava: typo

compiler-hacking.md
apmasell: It's not a special term, but it's something that is provided by `Input shesmu;`.

compiler-hacking.md
apmasell: I've rewritten this section.

compiler-hacking.md
```diff
@@ -0,0 +1,474 @@
+# Understanding the Shemsu compiler
+This is a guide to understanding the Shesmu olive compiler. This is meant to
+explain the core components that get the language to Java bytecode.
+
+It assumes you have a working understanding of Java bytecode. The [DZone
+bytecode
+tutorial](https://dzone.com/articles/nominalized-adjectives-as-names-for-decorators)
```
callunity: Is this link correct?

apmasell: No! And yet it is also the link in the implementation guide. :cry:

compiler-hacking.md
avarsava: Should this be 'when accessing the stream variable 'changed'', then?

compiler-hacking.md
```diff
@@ -0,0 +1,473 @@
...
+Shesmu has two kinds of stream purity: olive-level (type) and `For`-level
+(order). They are conceptually similar, but perform different functions.
+
+For all operations on Java streams, the order or type of the items can be
+changed by some operations.
+
+In the case of olives, `Group`, `Join`, `LeftJoin` change the type of the
+stream value while `Where`, `Monitor`, `Reject`, and `Pick` do not. A `Define`
+olive is written expecting the original type of the input stream. Therefore, no
+type-changing operation can precede a call, but the operations can. Therefore,
```
callunity: "Therefore, no type-changing operation can precede a call, but the operations can." I'm not quite understanding this...

compiler-hacking.md
```diff
@@ -0,0 +1,473 @@
...
+        INVOKEVIRTUAL dyn/shesmu/Program.clearGauge ()V
+
+Now there are interesting differences. First, the current time is stored for timing the olive runtime.
+
+       L0
+        INVOKESTATIC java/lang/System.nanoTime ()J
+        LSTORE 3
+
+Now, a stream of the correct type must be started that will contain the right
+objects. The supplied argument is `Function<Class<T>, Stream<T>>` and it's
```
callunity: `it's` -> `its`

compiler-hacking.md
```diff
@@ -0,0 +1,473 @@
...
+purpose is to provide a stream of values for a class. The class is part of the
+input format. For `Input shesmu;`, an `InputFormatDefinition` will have
```
callunity: I got a bit tripped up trying to keep all the Shesmu-specific terms in mental RAM, and I think a few links could help. Could you please link `input format` (in `The class is part of the input format.`) to [implementation.md](https://github.com/oicr-gsi/shesmu/blob/master/implementation.md#input-format)?

I also got a bit caught up on never having seen `Input shesmu;` in an olive before. It looks like that's the input format that Shesmu has built-in; would it be accurate to reword that sentence to the following?
An olive using `Input shesmu;` uses the `InputFormatDefinition` from [this file](https://github.com/oicr-gsi/shesmu/blob/master/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/core/input/shesmu/ShesmuIntrospectionFormatDefinition.java), which specifies that the class is `ShesmuIntrospectionValue`.

apmasell: I've rewritten this section to tease apart the input format details.

compiler-hacking.md
apmasell: Yeah, that was missing some adjectives. :dizzy_face:

compiler-hacking.md
```diff
@@ -0,0 +1,502 @@
...
+
+There are a lot of classes in the compiler, but the major ones are:
+
+- [`OliveNode`](shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/OliveNode.java), which declares each top-level element in the file (function, olive, define olive, constant)
+- [`OliveClauseNode`](shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/OliveClauseNode.java), which defines clauses in an olive
+- [`ExpressionNode`](shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/ExpressionNode.java), which defines all the bits and pieces of expressions
+
+For details on components, the JavaDoc outlines the behaviour and expectations
+of the classes. Most of the documentation is on the abstract classes and the
+concrete class have less since they just execute the behaviour defined by the
```
callunity: `class` -> `classes`

compiler-hacking.md
```diff
@@ -0,0 +1,502 @@
...
+abstract class.
+
+For general naming conventions:
+- _resolve_ methods convert strings into useful metadata (_e.g._ find the [`FunctionDefinition`](shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/FunctionDefinition.java) given the name in the script)
+- _render_ methods generate bytecode
+- classes that end in _Node_ are generate type of syntax in the language; their subclasses represent concrete bits of syntax the user can write
```
callunity: `are generate type of syntax in the language`?

compiler-hacking.md
```diff
@@ -0,0 +1,502 @@
...
+- classes that end in _Builder_ are utility classes that make generating bytecode simpler
+
+`ExpressionNode` has many similar clones. For instance, there is an expression
+node for `For` and then ones for each of the sources, clauses, subsamples, and
+collectors in a `For` expression, containing other expressions.  While these
+are structurally different, they are built very similarly to `ExpressionNode`.
+
+### Type Information
+Computing captures is done as part of bytecode production. The type system for
```
callunity: `Computing captures is done as part of bytecode production` <- feels kinda disjointed from both the section header and the next sentences, could it be given more context or moved?

compiler-hacking.md
```diff
@@ -0,0 +1,502 @@
...
+`path` type and the `+` operator can combine two paths. Shesmu will simply give
+up. This is meant to keep the compiler simple.
+
+### Bytecode Generation
+When generating bytecode, Shesmu uses
+[`GeneratorAdapter`](https://asm.ow2.io/javadoc/org/objectweb/asm/commons/GeneratorAdapter.html)
+from the ASM package to write bytecode. It wraps this in a
+[`Renderer`](shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/Renderer.java)
+class. `Renderer` provides a number of utility functions, such as inserting
+`Imyhat` objects into the bytecode. It's real purpose is to track extra
```
callunity: `It's` -> `its`

compiler-hacking.md
```diff
@@ -0,0 +1,502 @@
...
+(order). They are conceptually similar, but perform different functions.
+
+For all operations on Java streams, the order or type of the items can be
+changed by some operations.
+
+In the case of olives, `Group`, `Join`, `LeftJoin` change the type of the
+stream value while `Where`, `Monitor`, `Reject`, and `Pick` do not. A `Define`
+olive is written expecting the original type of the input stream. Therefore, no
+type-changing operation can precede a call, but the operations that do not
+modify the type can precede a call, including call operations to `Define`
```
callunity: `call operations` -> `calls`?

compiler-hacking.md
```diff
@@ -0,0 +1,502 @@
...
+### Stream Purity
+Shesmu has two kinds of stream purity: olive-level (type) and `For`-level
+(order). They are conceptually similar, but perform different functions.
+
+For all operations on Java streams, the order or type of the items can be
+changed by some operations.
+
+In the case of olives, `Group`, `Join`, `LeftJoin` change the type of the
+stream value while `Where`, `Monitor`, `Reject`, and `Pick` do not. A `Define`
+olive is written expecting the original type of the input stream. Therefore, no
```
callunity: How about "A `Define` olive has more stringent requirements; the original type of the input stream (as defined in `Input x;`) is the only valid input type for `Define` olives. Therefore, no [...]"

compiler-hacking.md
```diff
@@ -0,0 +1,502 @@
...
+type-changing operation can precede a call, but the operations that do not
+modify the type can precede a call, including call operations to `Define`
+olives that do not modify the type. Therefore, Shesmu must figure out a stream
+purity to ensure that call operations will be type-safe. Additionally,
+signatures must be computed over the clauses in the original type-pure stream
+plus the input half of the first transformative clause.
+
+In `For` clauses, the type is not important, however, the sorting is. In
+Shesmu, we want values given to olives to be stable. Therefore, if, say, a
```
callunity: Do you need this `Therefore`, since you've got one in the next sentence?

compiler-hacking.md
```diff
@@ -0,0 +1,502 @@
...
+meaningful operation. For instance, the following is not meaningful:
+
+    For x In xs: Reverse FixedConcat x With ","
+
+`Reverse` requires an order that isn't present in the original. The following is valid:
+
+    For x In xs: Sort x.timestamp Reverse FixedConcat x.library_name With ","
+
+Even though the final output doesn't contain the timestamps, the output will be
+stable based on the input.
```
callunity: Can you have a `For` with only order-destroying clauses within?

compiler-hacking.md
```diff
@@ -0,0 +1,577 @@
...
+        INVOKEINTERFACE ca/on/oicr/gsi/shesmu/Signer.finish ()Ljava/lang/Object; (itf)
+        CHECKCAST java/lang/String
+        ARETURN
+        MAXSTACK = 0
+        MAXLOCALS = 1
+
+</details>
+
+Finally, the definition for the method that is converted to a `Consumer` and
+fed to the `forEach` call above. This method will either to produce an action
```
callunity: There's something wonky about the first sentence here...

compiler-hacking.md
```diff
@@ -0,0 +1,577 @@
...
+
+Even though the final output doesn't contain the timestamps, the output will be
+stable based on the input.
+
+The final collector matters. If the operation is `Count`, then order is
+irrelevant. The following is valid, though kind of pointless:
+
+    For x In xs: Sort x.timestamp Flatten (y In x.libraries) Count
+
+The `Sort` operation has it's order destroyed by the `Flatten` and `Count`
```
callunity: `it's` -> `its`, and it parses better if you add a comma after `Flatten`

PR #195, opened 2019-01-10
Ignore any runs without a valid path

apmasell:


plugin/niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/RemotePineryIUSRepository.java
```diff
@@ -52,7 +52,7 @@ private ItemCache(Path fileName) {
 					final Map<String, Integer> badSetCounts = new TreeMap<>();
 					final Map<String, String> runDirectories = new HashMap<>();
 					final Set<String> completeRuns = c.getSequencerRun().all().stream()//
-							.filter(run -> run.getState().equals("Completed"))//
+							.filter(run -> run.getState().equals("Completed") && !run.getRunDirectory().equals(""))//
```
djcooke: Should probably add a null check for future-proofing since the field is nullable

apmasell: Fixed, but this makes me sad.

PR #200, opened 2019-01-11
Precache input formats and run olives in parallel

apmasell:
And do not enforce a timeout when gathering data

### DIFF
--- a/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/CompiledGenerator.java
+++ b/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/CompiledGenerator.java
@@ -3,16 +3,16 @@ package ca.on.oicr.gsi.shesmu;
 import java.nio.file.Path;
 import java.util.Collections;
 import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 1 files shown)


morgantaschuk:
200 PRs! :100:

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/CompiledGenerator.java
```diff
@@ -147,66 +142,56 @@ public void errorHtml(SectionRenderer renderer) {
...
-					@Override
-					public String name() {
-						return format.name();
-					}
-
-					@Override
-					public void run() {
-						try {
-							INPUT_RECORDS.labels(format.name()).set(format.input(format.itemClass()).count());
+		// TODO: collect only the formats that olives are currently using
```
callunity: Is there a ticket for this `TODO` (ie. can it be removed from the code)?

apmasell: There's a ticket, but, I would leave it in...

PR #204, opened 2019-01-15
Create an editor to manipulate the maintenance schedules

apmasell:


maintenance-editor/src/main/java/ca/on/oicr/gsi/shesmu/maintenance/Editor.java
```diff
@@ -0,0 +1,196 @@
...
+    add(panel);
+    final JToolBar toolbar = new JToolBar();
+    panel.add(toolbar, BorderLayout.PAGE_START);
+    final JButton add = new JButton("Add");
+    toolbar.add(add);
+    final JButton delete = new JButton("Delete");
+    toolbar.add(delete);
+    final JButton reload = new JButton("Revert");
+    toolbar.add(reload);
+    final JButton save = new JButton(UIManager.getIcon("FileView.floppyDriveIcon"));
```
morgantaschuk: The icon is cute but I'm not sure why you need it? You could also use `+` `-` `<-` for the other buttons with appropriate mouseovers.

maintenance-editor/src/main/java/ca/on/oicr/gsi/shesmu/maintenance/Editor.java
```diff
@@ -0,0 +1,196 @@
...
+    save.setToolTipText("Save");
+    final DateTableModel model = new DateTableModel();
+    final JTable table =
+        new JTable(model) {
+          private static final long serialVersionUID = 1L;
+
+          public Component prepareRenderer(TableCellRenderer renderer, int row, int column) {
+            final JComponent c = (JComponent) super.prepareRenderer(renderer, row, column);
+            if (model.rows.get(row)[0].isAfter(model.rows.get(row)[1])) {
+              c.setBorder(BorderFactory.createMatteBorder(1, 1, 1, 1, Color.RED));
```
morgantaschuk: Why a border? This might make sense if the table got sorted in ascending order but it doesn't seem to be sorted at the moment.

maintenance-editor/src/main/java/ca/on/oicr/gsi/shesmu/maintenance/Editor.java
```diff
@@ -0,0 +1,196 @@
...
+    delete.addActionListener(
+        event -> {
+          Arrays.stream(table.getSelectedRows()).forEach(model.rows::remove);
+          model.fireTableDataChanged();
+        });
+    save.addActionListener(
+        event -> {
+          try (PrintWriter writer = new PrintWriter(file)) {
+            for (LocalDateTime[] row : model.rows) {
+              writer.write(toUtc(row[0]));
```
morgantaschuk: From what I've read about LocalDateTime, I'm not sure it takes DST into account. So are time changes still up to the user to configure?

maintenance-editor/src/main/java/ca/on/oicr/gsi/shesmu/maintenance/Editor.java
```diff
@@ -0,0 +1,196 @@
...
+        file = chooser.getSelectedFile();
+      } else {
+        return;
+      }
+    } else if (args.length == 1) {
+      file = new File(args[0]);
+    } else {
+      System.err.println("Usage: maintenance-editor [maintenance.tsv]");
+      System.exit(1);
+      return;
```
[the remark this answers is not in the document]
apmasell: The Java flow system doesn't know that you don't return from `System.exit`, so it says I have a bunch of undefined variables. :man_shrugging:

maintenance-editor/src/main/java/ca/on/oicr/gsi/shesmu/maintenance/Editor.java
apmasell: The border appears if the window is invalid. That is, if start date is after end date.

maintenance-editor/src/main/java/ca/on/oicr/gsi/shesmu/maintenance/Editor.java
apmasell: It should do timezone correction. I read the file in `ZonedDateTime`, which does the DST correction when I go to an from `LocalDateTime`.

apmasell:
I think I've fixed most of the problems except the highlighting. There's not a lot I can do about that. Whichever side is selected ignores the correct background colour even though I has a setting that is supposed to do that.

maintenance-editor/src/main/java/ca/on/oicr/gsi/shesmu/maintenance/Editor.java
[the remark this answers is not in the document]
morgantaschuk: Ha. 🤦‍♀️

PR #221, opened 2019-02-14
Starvation and Pinery fixes

apmasell:
Fixes three major problems:

- slow olives would starve the thread pool and the tasks meant to kill them
- change olive timeout to correct units
- move Pinery IUS provider method to a place where the annotation scanner will see it

### DIFF
--- a/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/ProgramNode.java
+++ b/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/ProgramNode.java
@@ -118,7 +118,7 @@ public class ProgramNode {
   }
 
   public int timeout() {
-    AtomicInteger timeout = new AtomicInteger(20);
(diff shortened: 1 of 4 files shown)


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/ProgramNode.java
```diff
@@ -118,7 +118,7 @@ public void render(RootBuilder builder) {
   }
 
   public int timeout() {
-    AtomicInteger timeout = new AtomicInteger(20);
+    AtomicInteger timeout = new AtomicInteger(20 * 600);
```
callunity: Am I misreading this, or should this be `20 * 60`?

apmasell: :man_facepalming: Fixed.

PR #226, opened 2019-03-01
Add action to create symlink via SFTP

apmasell:


plugin-sftp/src/main/java/ca/on/oicr/gsi/shesmu/sftp/SftpServer.java
```diff
@@ -67,8 +67,14 @@ public FileAttributeCache(Path fileName) {
     }
   }
 
+  private static final ObjectMapper MAPPER = new ObjectMapper();
   private static final FileAttributes NXFILE = new FileAttributes.Builder().withSize(-1).build();
-
+  private static final Counter sshErrors =
+      Counter.build(
+              "shesmu_sftp_symlink_errors",
+              "The number of error communicating with the SFTP server while trying to create a symlink.")
```
morgantaschuk: *errors

plugin-sftp/src/main/java/ca/on/oicr/gsi/shesmu/sftp/SftpServer.java
```diff
@@ -121,6 +132,51 @@ public void configuration(SectionRenderer renderer) throws XMLStreamException {
...
+      try {
+        final FileAttributes attributes = sftp.lstat(link);
+        // File exists and it is a symlink
+        if (attributes.getType() == FileMode.Type.SYMLINK) {
+          // It's what we want; done
+          if (sftp.readlink(link).equals(target)) {
+            return new Pair<>(ActionState.SUCCEEDED, false);
+          } else {
+            // It's pointing to something else; recreate it
+            sftp.rm(link);
```
morgantaschuk: This behaviour is different than ln -s and might be worth re-thinking. There's a possibility for it to completely trash the target symlink in an automated way.

morgantaschuk: The other potential situation where this happens is when you have two files named the same that are headed for the same destination, e.g. two bam files from the same sample by different aligners.

apmasell: Yeah, I was making it more like `ln -sf`. I'll change this to just be an error.

PR #228, opened 2019-03-08
Add a button to expand a time range in schedule editor

apmasell:


callunity:
I hadn't appreciated that "Expand" adds time on both the start and end of the window. It might be worth adding a line of text explaining that in the pane.

apmasell:
I fixed the layout and added some help.

PR #235, opened 2019-03-15
Clean up Niassa plugin

apmasell:
Major changes:

- group previous workflow runs to correctly match existing workflows
- use scheduler directly via Java instead of shelling out to SeqWare JAR
- IntelliJ-recommended fixes


plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/niassa/WorkflowAction.java
```diff
@@ -339,89 +281,45 @@ public final ActionState perform(ActionServices actionServices) {
...
+                    fileSwids)
+                .getReturnValue();
+        server.get().analysisCache().invalidate(workflowAccession);
+        WorkflowRunAttribute attribute = new WorkflowRunAttribute();
+        attribute.setTag("major_olive_version");
+        attribute.setValue(Long.toString(majorOliveVersion));
+        server.get().metadata().annotateWorkflowRun(runAccession, attribute, null);
+        success = runAccession != 0;
+      } catch (Exception e) {
+        // Suppress all the batshit crazy errors this thing can throw
```
morgantaschuk: :+1:

PR #246, opened 2019-04-16
Add a button to pause all actions from an olive

apmasell:
This was a request by Mike. Right now, I'm not planning to save that state of which olives are paused. We can re-evaluate later.


callunity:
I clicked "Pause Actions" and reloaded the page, and the paused state doesn't seem to have been preserved.

apmasell:
Fixed. This was due to a rounding problem with times.

PR #248, opened 2019-04-18
Allow removing pauses from dead olives

apmasell:
Once an olive has been updated or deleted, there might still be pauses
associated with it. This allows removing these pauses for which there is no
olive on the dashboard.

### DIFF
--- a/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
+++ b/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
@@ -325,10 +325,109 @@ public final class Server implements ServerConfig, ActionServices {
                                   });
                         });
                 writer.writeEndElement();
+                final Map<String, Instant> currentOlives =
+                    compiler
+                        .dashboard()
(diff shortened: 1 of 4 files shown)


apmasell:
I moved the warning up and changed the text to indicate it's the olive below. Should I move the top warning table above the table of contents? That could get hidden as the TOC gets bigger.

I don't think there will be a race, but I'm not sure on the scenario you mean. If you put in a pause, then get a new olive and that olive produces new actions, those will run even with the old pause. If there's an action that's shared, it will be paused until the old olive is resumed. The information about which olives created an action, the pauses, and the running of an action (including checking the pauses) are all stored in separate concurrent collections. So, you won't be able to guarantee the ordering, but it won't corrupt itself.

callunity:
The warning and changed text looks good. I don't have strong feelings about moving the top warning.

(Andre and I had a conversation about my ambiguously worded "race condition" concern, and came up with an idea for a "Purge Actions from this old olive"-type button to un-throttle the new olive)

apmasell:
Added the purge actions button.

PR #251, opened 2019-04-23
Improve UI

apmasell:
- Get a new colour palette
- Make buttons appear more pressable
- Use different colours for different functions
- Automatically add items selected from dropdowns on the action dashboard
- Improve the way action states are shown on the action dashboard
- Add watermarks that indicate state to actions tiles
- Show errors and outputs in popup dialogs instead of adjacent table cell entries
- Add close button to dialog boxes
- Improve _Alerts_ dashboard
- Add a throbber for all AJAX loading
- Make bytecode and JSON sections animate when collapsed/expanded

### DIFF
--- a/plugin-jira/pom.xml
+++ b/plugin-jira/pom.xml
@@ -16,7 +16,20 @@
   <repositories>
     <repository>
       <id>atlassian-public</id>
-      <url>https://maven.atlassian.com/content/repositories/atlassian-public/</url>
+      <url>https://packages.atlassian.com/content/repositories/atlassian-public/</url>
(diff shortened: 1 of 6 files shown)


callunity:
On the Action dashboard
  * the separator between all the <tr> elements makes it look like the dropdown menus have no labels (it's not clear that they belong to the row above)
  * "Automatically add items selected from dropdowns on the action dashboard" doesn't work if you want to add the first item; even if you open the dropdown menu and click it, nothing happens (you have to press the [+] button to add it)

PR #254, opened 2019-04-26
Allow saving searches in the action dashboard

apmasell:


README.md
```diff
@@ -216,3 +216,24 @@ To access data remotely, create a file ending in `.`_format_`-remote` as follows
...
+## Saved Searches
+Shesmu's _Actions_ dashboard provides a way to sift through the actions that
+olives have generated. It can be useful to save these searches. By clicking the
+_Save Search_, the search will be saved in the browser. They can be shared by
+clicking the clipboard icon beside a saved search to copy the search and then
+using the _Paste Search Filter_ button on the dashboard and pasting in the text
+copied.
+
+To go beyond person-to-person sharing, the search filter JSON, created by
+either clicking the clipboard icon beside a search or pressing the _Show
```
callunity: `pressing` -> `clicking` for consistency

README.md
```diff
@@ -216,3 +216,24 @@ To access data remotely, create a file ending in `.`_format_`-remote` as follows
...
+Request_ button, can be saved to a file ending in `.search` in the Shesmu
+configuration directory. The name of the file will be used the name of the
```
callunity: `used the name` -> `used as the name`?

README.md
```diff
@@ -216,3 +216,24 @@ To access data remotely, create a file ending in `.`_format_`-remote` as follows
...
+search.
+
+It is not recommended to save searches that reference a particular olive source
+location. Every time the file is updated or Shesmu is restarted, the olive's
+timestamp will be updated and the filter will no longer match. The `time`
+property in the filter can be changed to `null` to avoid this issue. Even if
+this were not the case, it is possible tat the olive will move around in the
```
callunity: `tat` -> `that`

iainrb:
- Is it possible to view saved searches without eg. copying them to the clipboard and pasting into a text editor?
- I tried entering some JSON, saving a search, and clicking the 'copy definition' button, but when I pasted the copied value I just got an empty list '[]'. Is this expected when running locally in Docker?
- The 'x' to delete a saved search appears on the far right of the browser window, whereas everything else of importance is on the left, so it took me a while to realise it was there. Would it be a good idea to limit the width of the "saved search" bar?
shesmu

apmasell:
The empty list thing was a bug. I've fixed that. I've added the button. I've set the menu width to be shorter. It still looks kind of lost, but the delete buttons are definitely attached. I don't really know if there's a better way to design that menu.

apmasell:
> In the "Saved Searches" tab it says "Click any cell or heading in summary tables to view matching results." but the saved searches don't look like they're in a table. Is it possible to have that text say something like "Click any saved search to run it"?

I've made that text only show up when you click the stats button, because that's what that was for originally.

> The buttons take up more space at the top of the page than would be my preference, and it pushes the rest of the information down. Would you consider giving them less padding, or was this a specific choice?

I made them smaller, but I do want them to be relatively big or the kind of get lost.

iainrb:
Looks good to me now.

PR #256, opened 2019-05-01
Search improvements

apmasell:


shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/action/Action.java
```diff
@@ -37,6 +39,18 @@ public void accepted() {}
   @Override
   public abstract boolean equals(Object other);
 
+  /**
+   * Get the time this action was last modified outside of Shesmu
+   *
+   * <p>Since actions are associated with an external entity of time kind, get most recent
```
callunity: `time kind`?

apmasell: `some kind`. I swear, I do not write documentation on my phone no matter how much it looks like it.

PR #261, opened 2019-05-06
Determine workflow run staleness

apmasell:
Select the least stale, newest workflow run and use that information to determine the action state.


apmasell:
Okay. Redesigned after talking to Mike. Hopefully more capable of handling the different cases we have.

plugin-niassa+pinery/README.md
```diff
@@ -74,17 +74,17 @@ format:
...
-parameters and how they should be available to olives. Each parameter is
-defined as follows:
+The `accession` is the workflow SWID. `type` describes how LIMS keys will be
+associated with the Niassa workflows (_e.g._, non-top-level,  BCL2FASTQ,
+CellRanger). For most workflows, this should be `"FILES"`. Currently,
+`"CELL_RANGER"` is the only other supported value. When finding existing
+workflow runs that match, the `accession` is used; any alternate accessions
+that should be considered equivalent can be put in the `previousAccessions`
+array. The `maxInFlight` is a best-effort attempt to limit the number of
+simultaneous workflow runs launch by this olive. The `services` list is the
```
morgantaschuk: 'launched'

plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/niassa/AnalysisState.java
```diff
@@ -63,28 +60,59 @@ public AnalysisState(Pair<Integer, Integer> accessions, List<AnalysisProvenance>
             .collect(Collectors.toCollection(TreeSet::new));
   }
 
-  public boolean compare(
+  public Pair<AnalysisComparison, AnalysisState> compare(
       LongStream workflowAccessions,
       String majorOliveVersion,
-      String fileAccessions,
-      List<? extends LimsKey> limsKeys) {
+      Set<Integer> inputFiles,
```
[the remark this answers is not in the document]
morgantaschuk: We talked and clarified that this is specifically for Niassa, so SWIDs.

plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/niassa/AnalysisState.java
```diff
@@ -63,28 +60,60 @@ public AnalysisState(Pair<Integer, Integer> accessions, List<AnalysisProvenance>
...
+  /** Check how much this analysis record matches the data provided */
+  public Pair<AnalysisComparison, AnalysisState> compare(
       LongStream workflowAccessions,
       String majorOliveVersion,
-      String fileAccessions,
-      List<? extends LimsKey> limsKeys) {
+      Set<Integer> inputFiles,
+      List<? extends LimsKey> inputLimsKeys) {
+    // If the WF, version or input files doesn't match, bail
     if (workflowAccessions.noneMatch(a -> workflowAccession == a)
```
morgantaschuk: I don't think this is correct behaviour. I think the the olive should filter equivalent workflows, and we should be left with an action for a workflow that should definitely happen as long as the other conditions are met. With this syntax, I think that you couldn't e.g. run Mutect workflow and Strelka workflow on the same input files.

morgantaschuk: wait, are these workflow accessions or workflow *run* accessions?

apmasell: These are workflow accessions, not workflow run accessions. Mutect and Strelka should have different WF accessions and no “previous” WF accessions should be shared between them.

PR #262, opened 2019-05-07
Keep menus open when clicked

apmasell:


apmasell:
It's not intentional. I've changed it so the menu will stay open once you select an item (any item). I also changed it so if you mouse over another add button, it will close the selected menu, even if you clicked it open.

callunity:
It looks like only the "Add" menus are 'sticky'; can this be added to the "Ago" and "type of text search" menus too?

apmasell:
Okay. It's consistent now.

callunity:
I dunno about "need", but it was unexpected to have different behaviour for different dropdowns!

PR #265, opened 2019-05-15
Update logo to new colour palette

apmasell:
If you have more general opinions on the logo, now is a good time.


avarsava:
I added a thin (0.2px) black (dark grey would work too) outline to the knife and felt it was easier to tell at a glance what I was looking at - the contrast between the red and grey is a little harsh otherwise

apmasell:
Updated.

PR #268, opened 2019-05-29
Better histograms

apmasell:
This makes the histograms into clickable tiles instead of tables. Because starting a test server doesn't have any useful data, I've included a PR that provides simulated data for the stats page. I will delete it before merging.

### DIFF
--- a/shesmu-server/pom.xml
+++ b/shesmu-server/pom.xml
@@ -117,6 +117,16 @@
       <resource>
         <directory>src/main/resources</directory>
         <filtering>true</filtering>
+        <includes>
+          <include>**/shesmu-build.properties</include>
+        </includes>
+      </resource>
+      <resource>
+        <directory>src/main/resources</directory>
+        <filtering>false</filtering>
+        <excludes>
+          <exclude>**/shesmu-build.properties</exclude>
+        </excludes>
       </resource>
     </resources>
     <plugins>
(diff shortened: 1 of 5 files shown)


iainrb:
- Clickable tiles are nifty. As a further enhancement, it would be good to have histogram labels/axes -- I don't find it very intuitive to click on the tiles and then subtract interval lengths in my head to work out when each window covers.
- Could it measure intervals in days+hours instead of weeks+days+hours? I'm not sure "2w+3d+4h" gains us anything over "17d+4h" except the need to do more arithmetic.
- Clearer separation between the tiles and the table above them would be nice. (See screenshot.) 
shesmu20190529

apmasell:
I made most of those changes. I think weeks is just a thing I've absorbed from Prometheus, but it's really not helpful.

Anyway, for the histogram titles...I'm not quite sure what to do. I could put them diagonally up (or down) from every line in between the columns of tiles. I'm wary of having enough space if I put them directly under the column lines horizontally. What do you think?

iainrb:
Latest changes look good thanks.

Re. titles, I'd just like to be able to read off the approximate boundaries between tiles on the horizontal axis. Diagonal labels sound good.

apmasell:
Okay. I've added some labels. The even thin themselves out when you make the window too small.

PR #270, opened 2019-05-31
Allow doing string expansion for 10X barcodes

apmasell:
There's a bug in the functions page that means this can't be tested with the UI. It's fixed in the histogram PR.
```
$ curl -d '{"name":"expand_index", "args": ["A"]}' localhost:8081/function
{"value":["A"]}
$ curl -d '{"name":"expand_index", "args": ["SI-NA-A9"]}' localhost:8081/function
{"value":["ACAACTTG","CTCCAACA","GAGTGCGT","TGTGTGAC"]}
```

### DIFF
--- a/plugin-tsv/src/main/java/ca/on/oicr/gsi/shesmu/tsv/StringExpandFile.java
+++ b/plugin-tsv/src/main/java/ca/on/oicr/gsi/shesmu/tsv/StringExpandFile.java
@@ -1,0 +1,71 @@
+package ca.on.oicr.gsi.shesmu.tsv;
+
+import ca.on.oicr.gsi.shesmu.plugin.Definer;
+import ca.on.oicr.gsi.shesmu.plugin.PluginFile;
+import ca.on.oicr.gsi.shesmu.plugin.functions.ShesmuMethod;
+import ca.on.oicr.gsi.shesmu.plugin.functions.ShesmuParameter;
+import ca.on.oicr.gsi.status.SectionRenderer;
+import io.prometheus.client.Gauge;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.*;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+class StringExpandFile extends PluginFile {
+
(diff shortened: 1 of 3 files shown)


plugin-tsv/src/main/java/ca/on/oicr/gsi/shesmu/tsv/StringExpandFile.java
```diff
@@ -0,0 +1,80 @@
...
+
+  public StringExpandFile(Path fileName, String instanceName, Definer<StringExpandFile> definer) {
+    super(fileName, instanceName);
+    this.definer = definer;
+  }
+
+  @ShesmuMethod(
+      type = "as",
+      description =
+          "Expand a string to a set of strings or a set of containing only the input string.")
```
callunity: `a set of containing` -> `a set containing`?

plugin-tsv/src/main/java/ca/on/oicr/gsi/shesmu/tsv/StringExpandFile.java
```diff
@@ -0,0 +1,80 @@
...
+  public Set<String> $(
+      @ShesmuParameter(
+              description =
+                  "The innput string to find the in the table; if it does not exist in the table, it is returned as a list containing the input")
```
callunity: `innput` -> `input`

plugin-tsv/src/main/java/ca/on/oicr/gsi/shesmu/tsv/StringExpandFile.java
```diff
@@ -0,0 +1,80 @@
...
+  public StringExpandFile(Path fileName, String instanceName, Definer<StringExpandFile> definer) {
+    super(fileName, instanceName);
+    this.definer = definer;
+  }
+
+  @ShesmuMethod(
+      type = "as",
+      description =
+          "Expand a string to a set of strings or a set of containing only the input string.")
+  public Set<String> $(
```
callunity: Why the dollar sign as the variable name?

apmasell: The `$` gets substituted to the main part of the file name. I've given the method a more sensible name and moved this name into the annotations.

PR #288, opened 2019-06-28
Create tags and descriptions for olives

apmasell:
This only adds them. Making them generally useful will be done separately.


apmasell:
Yeah, it's looks awful. As far as I'm concerned, redesigning the dashboard is the next thing to do, so I'm just shoving the information in for now. I didn't add any search support to the action dash pending that redesign.

PR #289, opened 2019-07-02
Allow plugins to throttle an olive based

apmasell:
This allows plugins to specify services that their functions, constants, and
signatures require and then causes the olive to check if any of those are
throttled before it starts execution.

- Adds a method get the services for each plugin file
- Includes file name information in signature definitions
- Add services to our existing plugins
- Threads plugin information through the parse tree
- Adds a guard at the start of the olive that converts plugin file names to services and checks them


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/plugins/PluginManager.java
```diff
@@ -1130,6 +1155,33 @@ public static CallSite bootstrap(
         MethodHandles.foldArguments(methodHandle, instance.dynamicInvoker()));
   }
 
+  /** Bootstrap method to get a plugin collection as a stream of throttleable services */
+  public static CallSite bootstrapServices(
+      Lookup lookup, String methodName, MethodType methodType, String... fileNames) {
+    // Our goal here is to create  list of services that plugins use so we can block an olive if
+    // that happens. We're given the file names of those plugins as parameters to this method and
```
callunity: `if that happens` what is `that` in this context?

apmasell: Reworded.

PR #311, opened 2019-07-23
Fix purge button on Actions page

apmasell:


shesmu-server/src/main/resources/ca/on/oicr/gsi/shesmu/shesmu.js
```diff
@@ -1213,13 +1221,13 @@ function setColorIntensity(element, value, maximum) {
...
-function purge(filters) {
-  const targetElement = document.getElementById("results");
+function purge(filters, afterClose) {
+  const targetElement = makePopup(false, afterClose);
   if (filters.length == 0) {
     clearChildren(targetElement);
     const sarcasm = document.createElement("P");
     sarcasm.innerText =
-      "Yeah, no. You can't nuke all the actions. Maybe try a subset.";
+      "Yeah, no. You probably shouldn't nuke all the actions. Maybe try a subset.";
```
callunity: Can you make an additional button here for something like "Never mind, I'll filter first then nuke a subset"? The tiny "x" at the corner isn't nearly as attention-grabbing as the YOLO button.

apmasell: Now, with more sarcasm!

PR #328, opened 2019-08-08
Create Single collector for For expressions in addition to Group

apmasell:


language.md
```diff
@@ -1023,6 +1023,13 @@ accumulator, which will be returned, which is initially set to _initialexpr_.
 For every item, _expr_ is evaluated with _a_ set to the previously returned
 value.
 
+#### Single Value
+- `Single` _expr_ `Default` _defaultexpr_
+
+Evaluates all _expr_ for each item in the list and returns it if all are the same.
```
callunity: Looking at `Single`, I'd expect it to behave something like `First` (but somehow different), but that's not the case. What do you think about renaming it to `Same` or `IfSame`?

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/JavaStreamBuilder.java
```diff
@@ -329,6 +333,26 @@ public void reverse() {
     renderer.methodGen().invokeStatic(A_RUNTIME_SUPPORT_TYPE, METHOD_STREAM__REVERSE);
   }
 
+  public Renderer single(
+      int line, int column, LoadableConstructor name, LoadableValue... capturedVariables) {
+    finish();
+    final LambdaBuilder builder =
+        new LambdaBuilder(
+            owner,
+            String.format("For ⋯ First Default %d:%d", line, column),
```
callunity: Should `First` -> `Single`?

language.md
apmasell: Yeah, I'm not sure what to call it. I don't like `IfSame` because there's no if/else structure. Looking around:

- `Univalued`
- `Always`

There's already a `Single` in the grouping operation, so we should rename both since they do basically the same thing.

callunity: I like `Univalued`, and could be happy with `Always`

PR #337, opened 2019-08-20
Replace ternary if with If-Then-Else

apmasell:


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/ExpressionNode.java
```diff
@@ -309,7 +207,144 @@ private static Parser parse9(Parser input, Consumer<ExpressionNode> output) {
...
+                    p.column(),
+                    name.get(),
+                    source.get(),
+                    transforms.get(),
+                    collector.get()));
+          }
+          return result;
+        });
+    OUTER.addRaw(
+        "tenary if",
```
callunity: `tenary if`? (Probably doesn't need to be changed if this string is only exposed to/consumed by Shesmu, since it'll be deleted later)

apmasell: Fixed. It will appear in error messages.

apmasell:
`Switch` is already there. I've changed the section name to encompass everything in it now.

PR #347, opened 2019-08-28
Add Nothing keyword to create empty optional

apmasell:


callunity:
There's already the `nothing` action, and I wonder if these could be confused. Would `Empty` work as an alternative name for the empty optional?

apmasell:
Oh, yikes. I forgot. Internally, empty generally refers to an empty list. I think I'd rather rename the `nothing` action to `do_nothing`, `inaction` or something like that.

callunity:
Renaming makes sense to me. I'm a fan of `do_nothing`

apmasell:
To match the rest of the JSON action names with dashes. :sweat:

PR #356, opened 2019-09-10
Save selected olive in dashboard URL

apmasell:


shesmu-server/src/main/resources/ca/on/oicr/gsi/shesmu/shesmu.js
```diff
@@ -716,12 +716,28 @@ function initialise() {
...
 
-export function initialiseOliveDash(oliveFiles, deadPauses) {
+export function initialiseOliveDash(oliveFiles, deadPauses, saved) {
   initialise();
   const container = document.getElementById("olives");
   const results = document.getElementById("results");
 
   const renderOlive = (file, olive, pauseSpan) => {
+    if (window.history.state != name) {
+      window.history.pushState(
```
callunity: It looks like this isn't getting updated when a new olive is selected from the dropdown menu. I loaded the page, selected an olive from the dropdown, and the URL got updated. I then selected another olive from the dropdown, copied the URL and opened it in a new tab, and found it opened the first olive I had selected.

apmasell: Fixed, but kind of even more terrible.

PR #394, opened 2019-10-10
Show different icons for olive types

apmasell:


apmasell:
Does it make sense to have a single Olives tab that looks more like the olive dashboard with a dropdown that shows the diagrams?

callunity:
I like the idea of reusing pieces of the interface, but I think it would be important to keep the visual consistency between pages and keep it tabbed rather than in a dropdown. That would result in nesting tabs though.

PR #406, opened 2019-10-23
Display Cromwell ID for Guanyin Report actions

callunity:


plugin-guanyin/src/main/java/ca/on/oicr/gsi/shesmu/guanyin/GuanyinReportPluginType.java
```diff
@@ -13,7 +13,7 @@
   @Override
   public void writeJavaScriptRenderer(PrintStream writer) {
     writer.print(
-        "actionRender.set('guanyin-report', a => [title(a, `${a.reportName} – 观音 Report ${a.reportId}`)].concat(jsonParameters(a)).concat(a.cromwellUrl ? [link(a.cromwellUrl, 'Cromwell')] : []));");
+        "actionRender.set('guanyin-report', a => [title(a, `${a.reportName} – 观音 Report ${a.reportId}`)].concat(jsonParameters(a)).concat(a.cromwellUrl && a.cromwellId ? [link(a.cromwellUrl, 'Cromwell ID: ' + cromwellId)] : []));");
```
apmasell: Needs to be `a.cromwellId`.

callunity: So it does. Fixed.

PR #410, opened 2019-10-25
Improve import/export UI experience

apmasell:


apmasell:
Okay. There are two problems:

- pasting from the clipboard doesn't work as advertised because you need to enable some bit that you trust clipboard reads
- the import button was only for collections of searches, rather than individual searches. I've modified it to be able to handle both cases.

callunity:
I clicked "Save Search" and named it custom-ly, but nothing saved to disk. I then tried to "Import Searches" and there was no file to upload, so I clicked-and-pasted and got "Unexpected end of JSON input".

I was a bit surprised that the "click here and paste" didn't actually show me what was being pasted, or have any UI change after clicking. I was kinda expecting a text box or something; I worry that users will pause after clicking since nothing has changed, instead of trusting the instructions on the box. (I am often that kind of user...) What do you think about having the "click" make a text box appear, and then add an onpaste handler to obscure the data that gets pasted?

apmasell:
Ohhh. Mmm, _Save Search_ saves it to your browser's local state and adds it to the drop down. Maybe _Add to My Searches_?

As for the paste thing, I think I can do something better without a box.

callunity:
I like _Add to My Searches_! Can we label that dropdown? Right now it's displaying in my browser as "All Actions" which would be confusing.

callunity:
(Sorry, didn't mean to close the PR...)

apmasell:
Updated.

PR #413, opened 2019-10-28
Remove invalid characters from file names

apmasell:


callunity:
Oh no. What got put into the file name?

apmasell:
`miso_v2.2` was there. It's not so great.

PR #414, opened 2019-10-28
Collect and show errors on actions

apmasell:


plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/niassa/WorkflowAction.java
```diff
@@ -186,12 +189,15 @@ public final ActionState perform(ActionServices actionServices) {
...
       if (hasLaunched) {
+        this.errors =
+            Collections.singletonList("Workflow has already been attempted. Purge to try again.");
         return ActionState.FAILED;
       }
 
       // Check if there are already too many copies of this workflow running; if so, wait until
       // later.
       if (server.get().maxInFlight(workflowAccession)) {
+        this.errors = Collections.singletonList("Too many workflows running.");
```
callunity: Could this be clarified to indicate that this will resolve without user action? `Workflow has not yet launched because too many copies of this workflow are running. New workflows will be launched as current workflows complete.` or similar?

plugin-guanyin/src/main/java/ca/on/oicr/gsi/shesmu/guanyin/RunReport.java
```diff
@@ -268,12 +266,18 @@ public ActionState perform(ActionServices services) {
                   "1.0",
                   MAPPER.writeValueAsString(labels),
                   null);
+          this.errors = Collections.emptyList();
         } else if (cromwellId != null) {
           cromwellId = wfApi.status("v1", cromwellId.getId());
+          this.errors = Collections.emptyList();
+        } else {
+          this.errors =
+              Collections.singletonList("Report has already been launched but ID is unknown.");
```
callunity: Is it the case that this may resolve with time? It's not clear if user action is required, or what user action could be performed.

apmasell:
Updated.

PR #417, opened 2019-10-31
Add attributes to Pinery_IUS

callunity:
GLT-3048
Add: DV200, RIN, Run Status, UMIs

### DIFF
--- a/plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
+++ b/plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
@@ -152,6 +152,8 @@ public class PinerySource extends JsonPluginFile<PineryConfiguration> {
                             : lp.getCreatedDate().toInstant(),
                         basesMask,
                         lp.getSequencerRunPlatformModel(),
+                        Optional.empty(),
+                        Optional.empty(),
                         false);
 
                 if (badSetInRecord.isEmpty()) {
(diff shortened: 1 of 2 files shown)


plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
```diff
@@ -122,6 +122,21 @@ private ItemCache(Path fileName) {
                     lp.getSequencerRunName(), new Pair<>(runDirectory, basesMask));
                 final Instant lastModified =
                     lp.getLastModified() == null ? Instant.EPOCH : lp.getLastModified().toInstant();
+                String runStatus = "Unknown";
```
apmasell: Sadly, doing this will kill Pinery. Right now, I've got a  `Map<String, Pair<String, String>> runDirectoriesAndMasks` for the information that needs to go from lanes to samples. Could you move this outside the loop and build a `Map<String, RunDto>` and then pick form the map?

callunity: I've made some changes, can you let me know if that's roughly what you were thinking?

apmasell: That's exactly what I was thinking. :+1:

callunity: Documentation for 404 is added

PR #421, opened 2019-11-01
Add PineryIUS fields for Guppy

callunity:
GR-823 & GLT-3023
Waiting on a Pinery and then pipedev release (due November 11).

### DIFF
--- a/plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PineryIUSValue.java
+++ b/plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PineryIUSValue.java
@@ -11,6 +11,7 @@ import java.util.Optional;
 public final class PineryIUSValue {
   private final String bases_mask;
   private final Instant completed_date;
+  private final Optional<String> container_model;
   private final String donor;
   private final Optional<Double> dv200;
   private final String group_desc;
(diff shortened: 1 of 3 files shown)


callunity:
Tests will fail until Pinery is released with the relevant changes.

apmasell:
Yes, but the code changes look good.

callunity:
Great, just wanted to make it clear for the second reviewer :)

plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PineryIUSValue.java
```diff
@@ -140,6 +152,7 @@ public boolean equals(Object o) {
         && project.equals(that.project)
         && rin == that.rin
         && run_status.equals(that.run_status)
+        && sequencing_kit.equals(that.sequencing_kit)
```
djcooke: Could be null

callunity: Thanks. I've updated them to be Optionals to account for that.

callunity:
Right now they've got empty string as the default, so nothing will blow up. I don't currently have the capacity to fix it here and update the olives which use these fields, but will try to get it in before the Pinery release on Monday!

PR #426, opened 2019-11-04
Create an input format for Guanyin records

apmasell:


callunity:
^ enthusiasm as a result of GitHub having a bad time

PR #431, opened 2019-11-11
Fix post-Let gang typing problem

apmasell:
This fixes a bug where gangs were being type checked before type checking
should have commenced. It also adds gang support to `Let` to allow copying an
entire gang to a later step.

### DIFF
--- a/language.md
+++ b/language.md
@@ -691,10 +691,19 @@ The gang can then be used in the `By` of either a `Group` or `Pick` operation:
         paths = paths;
 
 The gang, in this case, `patient_tissue_prep` defines a number of fields that
-can be grouped together in `By` clause, prefixed with an `@`. The gang can
-also be converted to an underscore delimited string:
+can be grouped together in `By` clause, prefixed with an `@`.
+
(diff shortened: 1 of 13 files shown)


language.md
```diff
@@ -691,10 +691,16 @@ The gang can then be used in the `By` of either a `Group` or `Pick` operation:
...
+
+A gang can also be converted to an underscore delimited string:
 `"{@patient_tissue_prep}"`; the order of fields is defined by the input format.
-Additionally, the gang can be converted to a tuple `{@patient_tissue_prep}`.
+
+A gang can be converted to a tuple `{@patient_tissue_prep}`.
+
+A gang can also be used in a `Let` clause to preserve the contents of the gang:
+
+    Let timestamp = timestamp, @patient_tissue_prep
```
callunity: I find the syntax is a bit ambiguous here since it's all on one line. Is it equivalent to the following?
```
Let
    timestamp = timestamp,
    @patient_tissue_prep,
    project = project
```

apmasell: Yes. Updated.

PR #443, opened 2019-11-19
Allow incomplete runs in pinery_ius

apmasell:


plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
```diff
@@ -290,17 +279,11 @@ public ProjectCache(Path fileName) {
...
-    return val.orElse("");
-  }
-
-  private Optional<String> maybeGetRunField(RunDto run, Function<RunDto, String> getField) {
-    Optional<String> maybeVal = Optional.empty();
-    if (run != null) {
-      maybeVal = Optional.ofNullable(getField.apply(run));
-    }
-    return maybeVal;
+  private static boolean isRunComplete(RunDto run) {
```
callunity: Can this be renamed since it's no longer testing for the Completed status?

apmasell: Fixed.

PR #444, opened 2019-11-19
Add support for local variables

apmasell:


language.md
```diff
@@ -772,6 +772,13 @@ will produce:
 | `"a"`  | `` `1` `` | `` `True` `` |
 | `"b"`  | `` `2` `` | `` ` ` ``    |
 
+### Blocks
+- `Begin` _name0_ `=` _expr0_`;`[ _name1_ `=` _expr1_`;`[ ...]] `Return` _expr_`; End`
```
callunity: Since people will pattern their usage off these docs, can this be displayed in a code block (better readability)? Something like this?
```
Begin
  name0 = expr0;
  name1 = expr1;
  name2 = {name0, name1};
  [...];
  expr = List {name0, name1, ...};
  Return expr;
End

```

apmasell: Fixed. Markdown is a pain.

PR #462, opened 2019-12-03
Add sequencer start date

apmasell:
I'm assuming `startDate` is always in the usual Java zoned date time format.

### DIFF
--- a/plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
+++ b/plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
@@ -24,6 +24,7 @@ import io.prometheus.client.Gauge;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.time.Instant;
+import java.time.ZonedDateTime;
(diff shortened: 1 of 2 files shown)


plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
```diff
@@ -154,6 +155,9 @@ private ItemCache(Path fileName) {
                     maybeGetRunField(run, RunDto::getSequencingKit),
                     maybeGetRunField(run, RunDto::getContainerModel),
                     Optional.empty(),
+                    run.getStartDate() == null || run.getStartDate().isEmpty()
+                        ? Instant.EPOCH
+                        : ZonedDateTime.parse(run.getCreatedDate()).toInstant(),
```
djcooke: Looks like this is supposed to be `run.getStartDate()` rather than `run.getCreatedDate()`? Same below on L233

apmasell: Indeed. :dizzy_face:

PR #471, opened 2019-12-12
Include workflow SWIDs in text search

apmasell:


plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/niassa/WorkflowAction.java
```diff
@@ -345,6 +345,8 @@ public boolean search(Pattern query) {
     return ini.values().stream().anyMatch(v -> query.matcher(v.toString()).matches())
         || (runAccession != 0 && query.matcher(Integer.toString(runAccession)).matches())
         || limsKeysCollection.matches(query)
+        || workflowAccessions()
+            .anyMatch(workflow -> query.matcher(Long.toString(workflow)).matches())
```
callunity: (probably more accurate to say workflow run, but it's internal so doesn't really matter)

apmasell: No, this is actually the workflow SWID, not the run SWID.

callunity: ...so it is. My mistake!

PR #485, opened 2019-12-17
Allow throttling from olives and add user-defined frequency

asofmosq:
https://jira.oicr.on.ca/browse/GP-1893
https://jira.oicr.on.ca/browse/GP-1898

### DIFF
--- a/language.md
+++ b/language.md
@@ -497,6 +497,19 @@ Stop the script after _integer_ seconds. The usual integer suffixes, especially
 `minutes` and `hours` can be added here. When a script runs over its time
 budget, the Prometheus variable `shesmu_run_overtime` will be set. 
 
+### Frequency
+
+- `Frequency` _integer_ `;`
+
+Run the script every _integer_ seconds. The usual integer suffixes, especially `minutes`
+and `hours` can be added here. 
+
+###Required Services
+
+- `RequiredServices` _service1 [, service2, ...]_ `;`
+
+Calls the throttler to check for services. Multiple services are separated by a comma.
+
 ## Types
 There are a small number of types in the language, listed below. Each has
 syntax as it appears in the language and a descriptor that is used for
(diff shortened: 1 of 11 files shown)


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/PragmaNode.java
```diff
@@ -26,6 +29,43 @@
           }
           return result;
         });
+
+    DISPATCH.addKeyword(
+        "RequiredServices",
+        (input, output) -> {
+          final AtomicReference<List<String>> services =
+              new AtomicReference<>(Collections.emptyList());
```
apmasell: The empty list isn't necessary. If this doesn't get filled in, the code that uses it shouldn't be reached. No need to change it though.

language.md
```diff
@@ -497,6 +497,19 @@ Stop the script after _integer_ seconds. The usual integer suffixes, especially
 `minutes` and `hours` can be added here. When a script runs over its time
 budget, the Prometheus variable `shesmu_run_overtime` will be set. 
 
+### Frequency
+
+- `Frequency` _integer_ `;`
+
+Run the script every _integer_ seconds. The usual integer suffixes, especially `minutes`
+and `hours` can be added here.
```
callunity: This is a bit unclear - maybe indicate that it's `seconds` by default, but `minutes` and `hours` could be explicitly specified?

language.md
```diff
@@ -497,6 +497,19 @@ Stop the script after _integer_ seconds. The usual integer suffixes, especially
...
+- `Frequency` _integer_ `;`
+
+Run the script every _integer_ seconds. The usual integer suffixes, especially `minutes`
+and `hours` can be added here. 
+
+###Required Services
+
+- `RequiredServices` _service1 [, service2, ...]_ `;`
+
+Calls the throttler to check for services. Multiple services are separated by a comma.
```
callunity: I think that the concept of `the throttler` isn't really one that our users will have a solid mental model for. Maybe replace it with something like "Ensures that the olive will only run if the specified services are not throttled."?

PR #486, opened 2019-12-17
Alert dashboard

apmasell:


apmasell:
Also, the build is failing due to a function which is only in Java 11 that's in the commit to be removed.

PR #490, opened 2019-12-18
Allow back button and URL to restore state on action and olive dashbo…

apmasell:
…ards


apmasell:
I think it works now, maybe?

callunity:
Uhhh now the URL is not updating at all when I click on the top 5 olives in the dropdown, but it updates for all the ones below :sweat_smile:

apmasell:
Third times the charm?

PR #511, opened 2020-01-16
Allow copying of filters back to JSON

apmasell:


shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/filter/FilterBuilder.java
```diff
@@ -1,18 +1,148 @@
...
 import java.time.Instant;
 import java.util.List;
 import java.util.Optional;
-import java.util.function.Predicate;
 import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
 public interface FilterBuilder<F> {
+  /**
+   * Do a tranformation of the JSON representation to the same representation. This will make a copy
```
callunity: "the same representation"? If it's just copying, can you explain why?

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/filter/FilterBuilder.java
```diff
@@ -1,18 +1,148 @@
...
+        public FilterJson isState(ActionState... states) {
+          final FilterStatus result = new FilterStatus();
+          result.setState(states);
+          return result;
+        }
+
+        @Override
+        public FilterJson negate(FilterJson filter) {
+          final FilterJson copy = filter.convert(this);
+          copy.setNegate(!copy.isNegate());
```
callunity: As per our discussion: Thanks, I hate it. Please document :)

apmasell: Actually, it makes a copy, so it's safer than I remember. Hooray, (it/I) is not a total hack.

PR #513, opened 2020-01-16
Track source locations for alerts

apmasell:
This is the first in a larger set of changes.


callunity:
That's an unusually large diff; was Angie's filter code not autoformatted by the Maven plugin?

apmasell:
Sadly, IntelliJ does the sorting, not the presubmit hook. I've made another PR to include the sorting configuration for IntelliJ.

PR #522, opened 2020-01-24
Create plugin interface for searches

asofmosq:
+ some sorting in alphabetical order in PluginManager


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
```diff
@@ -468,6 +469,9 @@ public String activeUrl() {
               public Stream<Header> headers() {
                 final Instant now = Instant.now();
                 final ObjectNode searchInfo = RuntimeSupport.MAPPER.createObjectNode();
+                pluginManager
+                    .searches(FilterBuilder.JSON)
+                    .forEach(pair -> searchInfo.putPOJO(pair.first(), pair.second()));
```
apmasell: Fun bug while testing, the UI expects a list of filters. Can you make `pair.second()` into a singleton list?

apmasell: Actually, no, that fails to add the `type` attribute. This works:
```
.forEach(pair -> searchInfo.putArray(pair.first()).addPOJO(pair.second()));
```

apmasell:
No, that's my bug. I fixed it, but this probably needs a rebase.

PR #560, opened 2020-02-19
Improve scheduler

apmasell:
So, this is based on Mike's request. Currently, the scheduler takes all the actions, sorts them priority, and then checks each one on in sequence. This changes it in two ways:

* it creates a thread pool so that actions can run in parallel
* it takes all the actions, sorts them priority, filters out any that are currently running or have run recently, and selects the top _200 - currently running_ and throws them into the scheduler.

This is meant to make the scheduler starve low priority actions so new high priority actions can run when the system is overloaded.

### DIFF
--- a/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/ActionProcessor.java
+++ b/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/ActionProcessor.java
@@ -37,9 +37,8 @@ import java.time.Instant;
 import java.time.format.DateTimeFormatter;
 import java.util.*;
 import java.util.Map.Entry;
-import java.util.concurrent.ConcurrentHashMap;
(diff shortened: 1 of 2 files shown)


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/ActionProcessor.java
```diff
@@ -426,6 +426,16 @@ public String name(String input) {
       Gauge.build("shesmu_action_oldest_time", "The oldest action in a particular state.")
           .labelNames("state", "type")
           .register();
+  private static final Gauge scheduledInRound =
+      Gauge.build(
+              "shesmu_action_scheduled_in_round",
+              "The number of actions that where schedule for processing in the last round.")
```
callunity: `where` -> `were`

apmasell:
That's go to be from something else... This is running in dev okay.

PR #562, opened 2020-02-20
Handle concurrent Niassa cache population gracefully

apmasell:


plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/niassa/WorkflowAction.java
```diff
@@ -251,6 +254,12 @@ public final ActionState perform(ActionServices actionServices) {
                 .filter(pair -> pair.comparison() != AnalysisComparison.DIFFERENT)
                 .sorted()
                 .collect(Collectors.toList());
+      } catch (InitialCachePopulationException e) {
+        this.errors =
+            Collections.singletonList(
+                "Failed to pull previously workflow runs from Niassa. This can either be a temporary error if another action is fetching that data or a permanent one caused by an incorrect workflow SWID.");
```
iainrb: "previously" -> "previous"

PR #566, opened 2020-02-21
Replace loading animation

apmasell:


apmasell:
I wonder if the right answer is to add a delay and only display the throbber if, say, 500ms elapse.

callunity:
That would make sense to me!

apmasell:
Fixed!

PR #569, opened 2020-02-22
Add negated regular expression syntax

apmasell:


language.md
```diff
@@ -434,6 +434,11 @@ dates, the lesser value occurs temporally earlier.
 Check whether _expr_, which must be a string, must matches the provided regular
 expression.
 
+- _expr_` !~ /`_re_`/`
+
+Check whether _expr_, which must be a string, must not matches the provided
```
callunity: Perhaps `Check whether...must not matches...` -> ` Check that...does not match...`?

apmasell: Oh, this was some copy-pasta. The other one is also bad.

PR #573, opened 2020-02-25
Export new bases mask

apmasell:


shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/grouper/GrouperOutputs.java
```diff
@@ -49,6 +49,21 @@
       GrouperOutput<I, T> first, GrouperOutput<I, S> second) {
     return new GrouperOutputs<>(first, second);
   }
+  /**
+   * A grouper that exports two variables to the olive.
+   *
+   * @param first the type of the first variable that will be exported to the olive
+   * @param second the type of the second variable that will be exported to the olive
```
callunity: Looks like `@param third` should be described here?

apmasell: Ooops. Fixed.

PR #577, opened 2020-02-26
Add ... object and tuple element operators

apmasell:


language.md
```diff
@@ -575,12 +575,19 @@ Specifies a date and time. If the time is not specified, it is assumed to be mid
 Creates a new tuple with the elements as specified. The type of the tuple is
 determined based on the elements.
 
+Instead of expression, a `...`_expr_ can be used and this will expand all the
```
callunity: `Instead of expression` <- what does this mean?

apmasell: Err..I want to say "instead of an expression to create a single argument, a `...`_expr_ can be used to insert all the items in another tuple inline." Does that read better?

PR #597, opened 2020-03-09
Graphically show actions which are currently running

apmasell:


apmasell:
I'm really not seeing any slow down. I'm not sure what's going on.

asofmosq:
I also wasn't able to replicate the slow down.

PR #605, opened 2020-03-13
Export secondary naming scheme

apmasell:


plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
```diff
@@ -352,6 +352,18 @@ public PinerySource(Path fileName, String instanceName) {
...
   }
 
+  @ShesmuMethod(
+      name = "$_projects_with_secondary_scheme",
+      description =
+          "Projects marked with the secondary naming scheme from in Pinery defined in {file}.")
+  public Set<String> secondaryProjects() {
+    return projects
+        .get()
+        .filter(SampleProjectDto::isSecondaryNamingSCheme)
```
callunity: `SCheme` -> `Scheme`?

apmasell: Unfortunately, it is that way in Pinery and we didn't notice, so just rolling with it.

callunity: :grimacing: good job us

PR #609, opened 2020-03-19
SFTP fixes

apmasell:
I still don't know why authentication fails.


plugin-sftp/src/main/java/ca/on/oicr/gsi/shesmu/sftp/SftpServer.java
```diff
@@ -198,7 +198,13 @@ public synchronized void configuration(SectionRenderer renderer) throws XMLStrea
         // We've been told to blow it away
         if (force) {
           sftp.rm(linkStr);
-          sftp.symlink(linkStr, target);
+          // Fun fact: OpenSSH has these parameters reversed compared to the spec.
```
callunity: 😭

PR #638, opened 2020-04-03
Show errors in a table under the editor on the simulation dashboard

apmasell:


asofmosq:
Big fan of this addition 👍

apmasell:
Thanks. It was at Mei's request.

PR #703, opened 2020-04-28
Cancel the Niassa workflow when it is being skipped

apmasell:
This point was brought up by Mei when wondering if she could skip a running
workflow. It makes sense for Shesmu to tell Niassa to cancel the workflow.

### DIFF
--- a/plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/niassa/WorkflowAction.java
+++ b/plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/niassa/WorkflowAction.java
@@ -561,6 +561,14 @@ public final class WorkflowAction extends Action {
           attribute.setTag("skip");
           attribute.setValue("shesmu-ui");
           server.get().metadata().annotateWorkflowRun(runAccession, attribute, null);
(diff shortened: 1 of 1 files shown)


plugin-niassa+pinery/src/main/java/ca/on/oicr/gsi/shesmu/niassa/WorkflowAction.java
```diff
@@ -536,6 +536,14 @@ public synchronized boolean performCommand(String commandName) {
           attribute.setTag("skip");
           attribute.setValue("shesmu-ui");
           server.get().metadata().annotateWorkflowRun(runAccession, attribute, null);
+          final WorkflowRun run = server.get().metadata().getWorkflowRun(runAccession);
+          switch (run.getStatus()) {
+            case running:
+            case pending:
+            case submitted:
+              run.setStatus(WorkflowRunStatus.cancelled);
```
mlaszloffy, line 544: This probably should be submit_cancel to let the propagate statuses job kill the workflow run.

apmasell: Fixed. I really don't understand how these statuses work.

PR #739, opened 2020-05-22
Invalidate analysis cache after scheduling

apmasell:
This is a possible fix to duplicate action launching to ensure that fresh
records are seen immediately by other actions.


apmasell:
Mike thinks this will nuke Niassa.

PR #763, opened 2020-06-16
Create new front end infrastructure

apmasell:
Here's the base set of infrastructure changes. I will split up the rest of the changes I have to get out more reviewable pieces.

PR #764, opened 2020-06-16
New alert dashboard

apmasell:
This is both for review, but based on #763 as a test. It should probably get reviewed after that is merged.

### DIFF
--- a/shesmu-server-ui/src/alert.ts
+++ b/shesmu-server-ui/src/alert.ts
@@ -1,0 +1,690 @@
+import { fetchJsonWithBusyDialog, refreshable } from "./io.js";
+import {
+  StateSynchronizer,
+  tile,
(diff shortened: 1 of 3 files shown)


shesmu-server-ui/src/alert.ts
```diff
@@ -0,0 +1,676 @@
...
+                filterList.add({ type: "live", value: true, label: null });
+              }),
+              button("💤 Expired", "Not currently firing alerts.", () => {
+                close();
+                filterList.keepOnly((x) => x.type != "live");
+                filterList.add({ type: "live", value: false, label: null });
+              }),
+              createLabelFilter(
+                "🏷️ Has Label",
+                "Find actions a labels.",
```
avarsava, line 208: a

PR #780, opened 2020-07-02
Migrate simulation dashboard to new infrastructure.

apmasell:


shesmu-server-ui/src/simulation.ts
```diff
@@ -0,0 +1,781 @@
...
+                    .sort((a, b) => a.localeCompare(b))
+                    .map(
+                      (name) =>
+                        [name, (row: RefillerRecord) => row[name]] as [
+                          string,
+                          (row: RefillerRecord) => any
+                        ]
+                    )
+                )
+              : "No records provided to refiller."
```
avarsava, line 517: Should this be 'No records provided to refill'? or 'No records provided to _the_ refiller'?

shesmu-server-ui/src/simulation.ts
```diff
@@ -0,0 +1,781 @@
...
+                input[1],
+                ...Array.from(input[1][0].keys()).map(
+                  (i) =>
+                    [`Column ${i + 1}`, (row) => row[i]] as [
+                      string,
+                      (input: any[]) => any
+                    ]
+                )
+              )
+            : "No records provided to dumper."
```
avarsava, line 548: Same q as with refiller

shesmu-server-ui/src/simulation.ts
```diff
@@ -0,0 +1,781 @@
...
+        "Upload a file from your computer to simulate",
+        () =>
+          loadFile((name, data) => {
+            fileName = name;
+            editor.setValue(data, 0);
+          })
+      ),
+      buttonAccessory(
+        "🡇 Download File",
+        "Save this editor to your computer",
```
avarsava, line 646: Saving 'this editor' sounds a little like it'll download a whole executable or something, which I take it it's not doing?

apmasell: No, but that would be cool. ;-)

PR #787, opened 2020-07-07
Create user interface help/tips

apmasell:
This was an idea that Alexis had for the UI to help new users acclimate. I've put in a few bits of information so far, but I intend to solicit ideas from others.

### DIFF
--- a/shesmu-server-ui/src/help.ts
+++ b/shesmu-server-ui/src/help.ts
@@ -1,0 +1,269 @@
+import {
+  UIElement,
+  buttonIcon,
+  subscribedState,
+  addElements,
+  br,
+  blank,
+  hr,
+} from "./html.js";
+import { locallyStored } from "./io.js";
+import { Publisher, pubSubModel } from "./util.js";
+export type Area = "action" | "alert" | "olive" | "pauses" | "simulator";
+export type HotSpot = "stats-crosstab" | "stats-table" | "stats-histogram";
+/**
+ * Describes an individual help message that should appear.
+ */
+interface HelpTip {
+  /**
+   * A short (one line) description of the feature
+   */
+  summary: string;
+  /**
+   * A long description of the feature
+   */
+  description: string;
+  /**
+   * The hotspot (_i.e._, UI location) where this help tip should be displayed.
+   */
+  hotspot: HotSpot | null;
+  /**
+   * A general category for the kind of information in this tip so that everything in a category can be marked as read at once.
+   */
+  area: Area;
+}
+/**
+ * All the help for the whole UI. Each outer array represents a release version of Shesmu.
+ *
+ * Entries in the arrays can be replaced by null to hide this tip if the user upgrades past this version.
+ * This only makes sense after the fact. Suppose version 2 has a tip and then version 3 replaces that feature; replacing that tip with null will make it hidden from a user going directly from 1 to 3. It must not be deleted outright, since that will shuffle in indicies and that will display the wrong thing to the user.
+ */
+const tips: (HelpTip | null)[][] = [
+  [
+    {
+      summary: "Use advanced search for tag×tag filtering",
+      description:
+        "When using the basic search, multiple tags are treated as “any” matches, so selecting another tag will broaden rather than narrow the search. Switch to the advanced search to allow narrowing when searching tags.",
+      hotspot: "stats-crosstab",
+      area: "action",
+    },
+    {
+      summary: "Click table rows to filter",
+      description: "Click on table rows to drill down into those actions.",
+      hotspot: "stats-table",
+      area: "action",
+    },
+    {
+      summary: "Click and drag histogram to filter",
+      description:
+        "Clicking and dragging on a range will add a search filter to display only matching the actions in that range.",
+      hotspot: "stats-histogram",
+      area: "action",
+    },
+    {
+      summary: "Use the breakdown table to filter",
+      description:
+        "If there are many actions, a breakdown is provided with common alert attributes to allow fast filtering.",
+      hotspot: null,
+      area: "alert",
+    },
+    {
+      summary: "Filter olives with file names, description, and tags",
+      description:
+        "Use Description or Tag in the olive source to anotate olives (and the actions they generate). Those annotations and the file name can be used as search keywords in the olive selection box.",
+      hotspot: null,
+      area: "olive",
+    },
+    {
+      summary: "Custom actions can be uploaded",
+      description:
+        "The simulator has access to all the actions on a server, but the Extra Definitions tab allows uploading custom actions, for in-development workflows.",
+      hotspot: null,
+      area: "simulator",
+    },
+  ],
+];
+
+/**
+ * This is the help information stored in the browser to track what the user has seen.
+ */
+interface HelpState {
+  /** This is one above the highest version number the user has been exposted to. (This means it should be the length of the tips list once the system is initialised.)*/
+  version: number;
+  /**
+   * All the tips the user has yet to see. These are a pair of indices into the features list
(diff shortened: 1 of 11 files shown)


shesmu-server-ui/src/help.ts
```diff
@@ -0,0 +1,223 @@
...
+    {
+      summary: "Use advanced search for tag×tag filtering",
+      description:
+        "When using the basic search, multiple tags are treated as “any” matches, so selecting another tag will broaden rather than narrow the search. Switch to the advanced search to allow narrowing when searching tags.",
+      hotspot: "stats-crosstab",
+      area: "action",
+    },
+    {
+      summary: "Click table rows to filter",
+      description: "Click on table rows to drill drown into those actions.",
```
avarsava, line 46: drown -> down

shesmu-server-ui/src/help.ts
```diff
@@ -0,0 +1,223 @@
...
+      hotspot: "stats-table",
+      area: "action",
+    },
+    {
+      summary: "Click and drag histogram to filter",
+      description:
+        "Clicking and dragging on a range will filter only matching the actions.",
```
avarsava, line 53: This sounds like 'the filter will only be applied to the highlighted range', is that right?

shesmu-server-ui/src/help.ts
```diff
@@ -0,0 +1,223 @@
...
+   * The hotspot (_i.e._, UI location) where this help tip should be displayed.
+   */
+  hotspot: string | null;
+  /**
+   * A general category for the kind of information in this tip so that everything in a category can be marked as read at once.
+   */
+  area: Area;
+}
+/**
+ * All the help for the whole UI. Each outer array represents a release version of Shesmu.
```
avarsava, line 33: So as versions are released, this file will expand? Can't old tips be versioned like the rest of the code?
Also, why don't the text of the tips live with the definition of the widget? From a dev perspective it seems weird to me to have to find this file to look up the name given to the tip I want to display

shesmu-server-ui/src/help.ts
```diff
@@ -0,0 +1,223 @@
...
+        model.statusChanged({
+          version: input.version,
+          tips: input.tips.filter(
+            ([version, index]) =>
+              !predicate(version, index, tips[version]?.[index])
+          ),
+        });
+      }),
+      " | ",
+      buttonIcon("💯 Read All", "Mark all tips as read.", () => {
```
avarsava, line 131: I'm confused about the distinction between "mark all _these_ tips as read" vs "mark all tips as read"

shesmu-server-ui/src/help.ts
```diff
@@ -0,0 +1,223 @@
...
+          tips: input.tips.filter(
+            ([version, index]) =>
+              !predicate(version, index, tips[version]?.[index])
+          ),
+        });
+      }
+    });
+    return buttonIcon(
+      hasNew ? "ℹ️✨" : "ℹ️",
+      hasNew ? "New tips and available!" : "Previous tips and advice.",
```
avarsava, line 184: "New tips and available"?

shesmu-server-ui/src/help.ts
apmasell: So, in the usual tip systems, what I really want is "go die in a fire" button and I'm trying to phrase that delicately.

shesmu-server-ui/src/help.ts
apmasell: Yes, this will expand. If a user skips ahead multiple versions, I kind of want them to get all the tips, so that's why I had it live with the code. I'm not really sure how this works in other system.

I agree that it's awkward for the tips to not live with the widget, but I want to have the dashboard-level tips include the tips of their components, unless you think that's not worthwhile.

shesmu-server-ui/src/help.ts
avarsava: image
So on this page, there is only 1 tip, so I'd expect 'Read' and 'Read All' to do the same action. 'Read All' would actually mark off _all_ tips, in all the boxes across Shesmu, until the next version is released? In that case I'd recommend renaming Read All to something like 'Turn off help notifications' and maybe move it to the navbar or somewhere else

shesmu-server-ui/src/help.ts
avarsava: Okay, so only 'new' tips would appear in each outer array, not a whole copy of the entire tip set? 

If Feature A is released in version 2 and gets a tip, but then in version 3 Feature A is rewritten and the behaviour changes and gets another new tip, would a user skipping from version 1 to version 3 see the now-erroneous tip for version 2?

apmasell: Hmm. If we deleted the erroneous version 2 tip, then yes, but if it wasn't the last tip, that would probably break everything. Maybe the answer is to add a deprecated flag that can be switched on for older tips.

shesmu-server-ui/src/help.ts
apmasell: Okay, I'll make it visible on the main Shesmu status page only. I've upgraded the tooltips on read to provide some more context too.

shesmu-server-ui/src/help.ts
avarsava: That might be a good idea

apmasell: Done. You can replace the block with `null` and it should do the right thing.

shesmu-server-ui/src/help.ts
```diff
@@ -0,0 +1,253 @@
...
+          version: input.version,
+          tips: input.tips.filter(([version, index]) => {
+            const tip = tips[version]?.[index];
+            return tip && !predicate(version, index, tip);
+          }),
+        });
+      }
+    });
+    return buttonIcon(
+      hasNew ? "ℹ️✨" : "ℹ️",
```
mlaszloffy, line 201: ⓘ ?

shesmu-server-ui/src/help.ts
avarsava: Opening the tooltip on the main page then clicking away (without clicking Read or Read All) clears the notifications for all of shesmu. Also, the naming is still confusing because 'Read All' being _inside_ the tooltip implies those actions only apply within the context of that tooltip.

shesmu-server-ui/src/help.ts
avarsava: bump

avarsava:
image
Once read, the tips and their titles all have the same font, which makes distinguishing titles from tips confusing

avarsava:
The Olives page tooltip is empty but still has a 'Read' button which produces a Javascript error ('Node.removeChild: The node to be removed is not a child of this node') when clicked

shesmu-server-ui/src/help.ts
apmasell: Rewritten.

shesmu-server-ui/src/help.ts
apmasell: Yeah, that was a mis-feature by design. I've updated the code to track which have been read and then only clear those.

apmasell:
I have a fix for the JavaScript error.

apmasell:
I've added some unattractive separators between tips. I could also start each summary with a lightbulb emoji.

avarsava:
Lightbulb emoji sounds like a good idea.
glad the JS error is resolved, but still confused as to why there is a i icon with a 'read' button available at all if the Olives page has no hints (even previously-read ones) to show me. It looks as though it failed to load or something

apmasell:
No, that's just my having forgotten to write a tip. :man_facepalming:

PR #792, opened 2020-07-07
Allow locking out UI during certain server operations

apmasell:


shesmu-server-ui/src/io.ts
```diff
@@ -334,13 +336,17 @@ export function refreshableSvg<I>(
  *
  * @param input the request to make
  * @param model the model that will deal with the response
+ * @param modal display a modal dialog to lock out the whole UI
  */
 export function requestModel(
   input: RequestInfo,
-  model: StatefulModel<Promise<Response>>
+  model: StatefulModel<Promise<Response>>,
+  modal: boolean
```
avarsava, line 344: thank god for typescript or this would seem Perilous

apmasell: I hate it, but I don't have better words.

PR #801, opened 2020-07-09
Fix advanced search (and related display issues)

apmasell:
- Fix query time conversion
  Times were incorrectly converted to seconds instead of milliseconds since the epoch.
- Filter out times that are the initial default
   This removes them from histograms where they provide no useful information.
- Prevent time windows in histograms for being empty 
   This solves a problem where rounding can make the selected histogram window have no span, resulting in a confusingly empty window.
- Allow synchronizers to self-update
   This is allows a synchronizer to see changes made by itself. The `internal` flag allows operations to avoid infinite updating.
- Create a bypass model    
   This works somewhat opposite to a split model and allows a model to take a single input and decide between two paths.
- Make advanced search work
   This change correctly wires in the advanced search and allows proper switching between advanced and basic search.
- Avoid plastering error messages everywhere
   This allow some panels to display the dead olive rather than the error message itself.


avarsava:
image
display issue with AND filter button

avarsava:
'Fixed Time Range' filters don't do anything when I try to add them

apmasell:
Okay, The time range is thing is working as intended, but bad intentions. You can specify a start and/or end time range; if neither is specified (aka unbounded), this does nothing. Neither is the default. I've set a default to be the from -1hour to now and I show an error if both are unbounded.

As for the _And Filter_ button, it was supposed to be a heavy ampersand, but I guess that's not rendering. I'm changing the design of this so that you select the filter (using an emoji that renders better) and then, only if you have an existing query, it asks if you want to do an AND or an OR with the new filter.

shesmu-server-ui/src/stats.ts
```diff
@@ -218,7 +218,7 @@ function renderStat(
               addRangeSearch(
                 bin as TimeRangeType,
                 boundaries[start],
-                boundaries[end]
+                Math.max(boundaries[end], boundaries[start] + 60_000) // Time is rounded to minutes for some situations, so if the window is too small, the start and end times will bet the same and nothing matches; this kicks the end range a bit into the future
```
mlaszloffy, line 221: bet --> be at

PR #803, opened 2020-07-09
Update installation instructions and Dockerfile

apmasell:
This updates the installation instructions to match the new build process
required to prepare the TypeScript user interface and also makes the matching
changes to the Dockerfile.

### DIFF
--- a/shesmu-server/src/test/java/ca/on/oicr/gsi/shesmu/UserInterfaceTest.java
+++ b/shesmu-server/src/test/java/ca/on/oicr/gsi/shesmu/UserInterfaceTest.java
@@ -38,7 +38,16 @@ public class UserInterfaceTest {
 
     final ChromeOptions opts = new ChromeOptions();
     opts.setHeadless(true);
-    opts.addArguments("--disable-gpu");
+    opts.addArguments("--window-size=1920,1080");
+    // I would love to tell you what these do, but without them, you get a nonspecific ChromeDriver
+    // error under Docker.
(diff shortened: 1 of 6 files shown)


shesmu-server/src/test/java/ca/on/oicr/gsi/shesmu/UserInterfaceTest.java
```diff
@@ -38,7 +38,16 @@ public UserInterfaceTest() {
 
     final ChromeOptions opts = new ChromeOptions();
     opts.setHeadless(true);
-    opts.addArguments("--disable-gpu");
+    opts.addArguments("--window-size=1920,1080");
+    // I would love to tell you what these do, but without them, you get a nonspecific ChromeDriver
+    // error under Docker.
+    // https://stackoverflow.com/questions/50642308/webdriverexception-unknown-error-devtoolsactiveport-file-doesnt-exist-while-t/50642913#50642913
+    opts.addArguments("start-maximized"); // open Browser in maximized mode
+    opts.addArguments("disable-infobars"); // disabling infobars
```
avarsava, line 46: to confirm - these two don't use '--'?

apmasell: It was like that in the Stack Overflow. :man_shrugging: It doesn't produce any errors for me.

apmasell:
The `unknown error: DevToolsActivePort file doesn't exist` was the original one I had and what the changes from Dillan fixed for me. I'm really not sure what to do.

PR #826, opened 2020-07-21
Handle multiple tags in drilldown properly

apmasell:
- Fix filter type in crosstab cells
- Upgrade to advanced search when basic won't do  
    When multiple tags are selected, the drill-down option doesn't work because the
    tags are OR'd together rather than AND'd together. If the user attempts this,
    switch them to the _Advanced_ search interface.

### DIFF
--- a/shesmu-server-ui/src/actionfilters.ts
+++ b/shesmu-server-ui/src/actionfilters.ts
@@ -1608,8 +1608,12 @@ function searchBasic(
     },
 
     addPropertySearch: (...limits) => {
-      updateBasicQueryForPropertySearch(limits, current);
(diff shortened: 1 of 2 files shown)


avarsava:
My local shesmu doesn't have any data to actually display, so I can't properly test this. When it switches to Advanced view, does it warn the user and let them abort the drilldown?
(Just noticed that there's no warning when pressing the Advanced button, maybe there should be a 'this is a one-way road' warning there too)

apmasell:
It doesn't warn, but I can add that. You should be able to use the `demo` directory now to bring up a Shesmu with plenty of actions.

PR #840, opened 2020-08-05
Add search import button

apmasell:
This was accidentally removed during the UI redesign.

### DIFF
--- a/shesmu-server-ui/src/action.ts
+++ b/shesmu-server-ui/src/action.ts
@@ -17,6 +17,7 @@ import {
   historyState,
   img,
   inputText,
+  inputTextArea,
   link,
   makeUrl,
   multipaneState,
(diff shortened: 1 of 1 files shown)


avarsava:
I see there's an Upload button, but what do I upload? I can't seem to find an Export button

apmasell:
Oh, it's on the lower toolbar: _Export Search_ and then either _To Clipboard_, _To Clipboard for Ticket_, or _To File_ will be a format this can read. I will add some text for this though.

avarsava:
> Oh, it's on the lower toolbar: _Export Search_ and then either _To Clipboard_, _To Clipboard for Ticket_, or _To File_ will be a format this can read. I will add some text for this though.

Sorry, where is the lower toolbar? 
image

apmasell:
I think there's another bug. It only shows if there are any actions, which isn't quite right. I'll fix that.

apmasell:
Fixed. It should display now. I didn't notice because my testing instance has some fake actions in it.

avarsava:
image
It does display, but I think it's incorrect

avarsava:
Also, probably unrelated, but when I save a search, then leave Shesmu and come back, and select my search from the dropdown, the filters bar is empty, is that intentional?

apmasell:
Yes, that is intentional. The search is the selected search + additional filters, so once you save it, that's not visible.

PR #1008, opened 2021-02-24
Correct multiple bugs in Vidarr plugin

apmasell:


pruzanov (approved):

apmasell:
Peter, this has gone through some substantial changes as I did some more testing so please have  a second look.

asofmosq (approved):

PR #1031, opened 2021-03-16
Fix SSH connection pooling bug

apmasell:
In a recent misconfiguration, the SSH connection didn't connect upon
initialisation. This caused the max connection counter to become depleted and
subsequent connection attempts would wait forever for a connection. This
applies two fixes:

- it sets a maximum limit for how long a caller will wait to acquire a connection
- it release the max connection counter if an exception is thrown during initialisation
n# Please enter the commit message for your changes. Lines starting

### DIFF
--- a/plugin-sftp/src/main/java/ca/on/oicr/gsi/shesmu/sftp/SshConnectionPool.java
+++ b/plugin-sftp/src/main/java/ca/on/oicr/gsi/shesmu/sftp/SshConnectionPool.java
@@ -37,12 +37,17 @@ public class SshConnectionPool implements Supplier<PooledSshConnection>, AutoClo
 
     public PooledSshConnection(ConnectionInfo info) throws IOException {
       this.info = info;
-      client = new SSHClient();
-      client.addHostKeyVerifier(new PromiscuousVerifier());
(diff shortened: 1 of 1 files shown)


asofmosq (approved):

plugin-sftp/src/main/java/ca/on/oicr/gsi/shesmu/sftp/SshConnectionPool.java
```diff
@@ -111,12 +116,13 @@ public void configure(String host, int port, String user) {
 
   @Override
   public PooledSshConnection get() {
+    int attempts = 100;
```
pruzanov, line 119: a hundred? That seems a lot

pruzanov (approved):

plugin-sftp/src/main/java/ca/on/oicr/gsi/shesmu/sftp/SshConnectionPool.java
apmasell: :man_shrugging: It's less than infinity.

PR #1035, opened 2021-03-20
Upgrade to Java 14

apmasell:
* Change language version to Java 14
* Use new Java 14 language features
* Switch to JDK HTTP client


maintenance-editor/src/main/java/ca/on/oicr/gsi/shesmu/maintenance/Editor.java
```diff
@@ -58,7 +57,7 @@ public int getRowCount() {
 
     @Override
     public Object getValueAt(int row, int col) {
-      final LocalDateTime[] rowValues = rows.get(row);
+      final var rowValues = rows.get(row);
```
avarsava, line 60: imo not knowing rowValues is a LocalDateTime[] makes the following line much more confusing, but this is a complaint about 'var' in general

avarsava (approved):

maintenance-editor/src/main/java/ca/on/oicr/gsi/shesmu/maintenance/Editor.java
apmasell: Yeah. Are you using IntelliJ? It displays the inferred type in grey, which is annoying in its own way.

avarsava: I'm using the diff view on github :wink: But yeah, the intellij thing is also annoying. Is there a good reason for switching to var everywhere other than 'we might change the type down the road and don't want to change the term on this particular line'?

apmasell: I'm not even sure that's the point of `var`. I think it's to have the ergonomics of not having to redundantly write the same type information in two places, usually on the same line. I guess I don't spend much time caring what the type of a variable is in most contexts. In Shesmu, I tend to use a lot of very nested structures that have a lot of generic information and `<>` doesn't do a good enough job of not having to write redundant type information, unfortunately.

avarsava: If you're calling `new` i guess I can see that, but in this particular line the type information is written _nowhere_, and adds a bit more mental complexity to figuring out what the next line is doing, especially in this environment where the fact that it's inside the Date Table class is obscured
anyway +1 on this change anyway, i just dislike var lol

asofmosq (approved):

PR #1037, opened 2021-03-25
Niassa migration action

asofmosq:
Some changes were just the auto-formatting that happened with the commit.

### DIFF
--- a/plugin-niassa/src/main/java/ca/on/oicr/gsi/shesmu/niassa/CustomInputLimsKeyProvider.java
+++ b/plugin-niassa/src/main/java/ca/on/oicr/gsi/shesmu/niassa/CustomInputLimsKeyProvider.java
@@ -48,4 +48,16 @@ public class CustomInputLimsKeyProvider implements InputLimsKeyProvider {
       }
     };
   }
+
+  @Override
+  public CustomActionParameter<MigrationAction> parameterMigration() {
+    return new CustomActionParameter<MigrationAction>("wdl_outputs", true, type) {
+      @Override
+      public void store(MigrationAction action, Object value) {
+        final List<Pair<String, CustomLimsEntry>> entries = new ArrayList<>();
+        converter.write(type, value, (name, handler) -> entries.add(new Pair<>(name, handler)));
+        action.limsKeyCollection(new CustomLimsKeys(entries));
+      }
+    };
+  }
 }
--- a/plugin-niassa/src/main/java/ca/on/oicr/gsi/shesmu/niassa/MigrationAction.java
+++ b/plugin-niassa/src/main/java/ca/on/oicr/gsi/shesmu/niassa/MigrationAction.java
@@ -1,0 +1,405 @@
+package ca.on.oicr.gsi.shesmu.niassa;
+
+import ca.on.oicr.gsi.Pair;
+import ca.on.oicr.gsi.prometheus.LatencyHistogram;
+import ca.on.oicr.gsi.provenance.model.LimsKey;
+import ca.on.oicr.gsi.shesmu.plugin.Definer;
+import ca.on.oicr.gsi.shesmu.plugin.Utils;
+import ca.on.oicr.gsi.shesmu.plugin.action.Action;
+import ca.on.oicr.gsi.shesmu.plugin.action.ActionParameter;
+import ca.on.oicr.gsi.shesmu.plugin.action.ActionServices;
+import ca.on.oicr.gsi.shesmu.plugin.action.ActionState;
+import ca.on.oicr.gsi.shesmu.plugin.cache.InitialCachePopulationException;
+import ca.on.oicr.gsi.shesmu.vidarr.RunState;
+import ca.on.oicr.gsi.shesmu.vidarr.RunStateAttemptSubmit;
+import ca.on.oicr.gsi.vidarr.api.ExternalKey;
+import ca.on.oicr.gsi.vidarr.api.SubmitWorkflowRequest;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.sql.Connection;
+import java.sql.DriverManager;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.util.*;
+import java.util.List;
+import java.util.Map;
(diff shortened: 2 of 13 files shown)


plugin-niassa/src/main/java/ca/on/oicr/gsi/shesmu/niassa/CustomInputLimsKeyProvider.java
```diff
@@ -48,4 +48,15 @@ public void store(WorkflowAction action, Object value) {
       }
     };
   }
+
+  public CustomActionParameter<MigrationAction> parameterMigration() {
+    return new CustomActionParameter<MigrationAction>("wdl_outputs_migration", true, type) {
```
apmasell, line 53: I think this should stay `wdl_outputs`. It' used for generating the INI, so it might not matter at all.

plugin-niassa/src/main/java/ca/on/oicr/gsi/shesmu/niassa/NiassaServer.java
```diff
@@ -789,4 +790,9 @@ private void updateWorkflows() {
   public String url() {
     return url;
   }
+
+  public Optional<URI> vidarrUrl() {
+    String vidarrUrl = configuration.map(Configuration::getVidarrUrl).orElse(null);
+    return vidarrUrl != null ? Optional.of(URI.create(vidarrUrl)) : Optional.empty();
```
apmasell, line 796: More compactly:
```
return configuration.map(Configuration::getVidarrUrl).map(URI::create);
```
but I don't really care in this case since this isn't a maintenance concern.

plugin-niassa/src/main/java/ca/on/oicr/gsi/shesmu/niassa/WorkflowConfiguration.java
```diff
@@ -129,9 +230,11 @@ public String applyTuple(Stream<Field<Integer>> contents) {
   private long[] previousAccessions;
   private boolean relaunchFailedOnUpgrade;
   private List<String> services = Collections.emptyList();
-  private InputLimsKeyProvider type;
+  private InputLimsKeyProvider<WorkflowAction> type;
```
apmasell, line 233: I think it would be better to change the `InputLimsKeyProvider` to avoid having to duplicate this in all the `.niassawf` files. If you make it look like this:
```
public interface InputLimsKeyProvider {
  <T extends Action> CustomActionParameter<T> parameter();
}
```
Then the method will be generic without the interface being generic and the two can share an implementation.

plugin-niassa/src/main/java/ca/on/oicr/gsi/shesmu/niassa/MigrationAction.java
```diff
@@ -0,0 +1,287 @@
...
+
+    return workflowAccession == that.workflowAccession
+        && Objects.equals(annotations, that.annotations)
+        && Objects.equals(ini, that.ini)
+        && Objects.equals(limsKeysCollection, that.limsKeysCollection);
+  }
+
+  @Override
+  public void generateUUID(Consumer<byte[]> digest) {
+    digest.accept(Utils.toBytes(workflowAccession));
```
apmasell, line 108: These will have the same ids as the real actions, which...I'm not sure if that's a good idea or not. We should discuss that.

plugin-niassa/src/main/java/ca/on/oicr/gsi/shesmu/niassa/MigrationAction.java
```diff
@@ -0,0 +1,287 @@
...
+      e.printStackTrace();
+      return ActionState.FAILED;
+    }
+
+    if (matches.isEmpty()) {
+      // go into waiting state and try again later
+      this.errors = Collections.singletonList("No candidate workflow runs for migration.");
+      return ActionState.WAITING;
+    } else {
+      ObjectMapper MAPPER = new ObjectMapper();
```
apmasell, line 207: This can be a static field

plugin-niassa/src/main/java/ca/on/oicr/gsi/shesmu/niassa/MigrationAction.java
```diff
@@ -0,0 +1,287 @@
...
+      request.setConsumableResources(new TreeMap<>());
+      request.setTarget("MIGRATION");
+      request.setWorkflow(workflowName);
+      request.setWorkflowVersion(matches.get(0).state().workflowVersion());
+      request.setLabels(labels);
+
+      if (stale) {
+        return ActionState.ZOMBIE;
+      }
+      final var throttled = actionServices.isOverloaded(this.services);
```
apmasell, line 220: This can be dropped. We're not using the services for real, so it doesn't matter.

plugin-niassa/src/main/java/ca/on/oicr/gsi/shesmu/niassa/MigrationAction.java
```diff
@@ -0,0 +1,287 @@
...
+      ObjectMapper MAPPER = new ObjectMapper();
+      ObjectNode labels = MAPPER.createObjectNode();
+      annotations.forEach(labels::put);
+
+      request.setConsumableResources(new TreeMap<>());
+      request.setTarget("MIGRATION");
+      request.setWorkflow(workflowName);
+      request.setWorkflowVersion(matches.get(0).state().workflowVersion());
+      request.setLabels(labels);
+
```
apmasell, line 216: Okay, so TODO here is to populate arguments and metadata.

apmasell (commented):
This is looking really good!

avarsava (approved):

plugin-niassa/src/main/java/ca/on/oicr/gsi/shesmu/niassa/MigrationAction.java
```diff
@@ -0,0 +1,385 @@
...
+      // Populate metadata
+      final var outputFileMigration = MAPPER.createArrayNode();
+      final var outputFiles = matches.get(0).state().files();
+      outputFiles.forEach(
+          file -> {
+            var provider = file.iterator().next().first();
+            var id = file.iterator().next().second();
+            String tree =
+                "{\"fileSWID\": "
```
apmasell, line 292: I would use the `MAPPER` to create this programmatically. There can be multiple LIMS keys per file, so you'll have to construct an array of them.

apmasell (approved):

PR #1046, opened 2021-03-31
Allow setting a custom set of clinical projects in Pinery

apmasell:
The `isClinical` method is deprecated and this is the replacement.


djcooke (approved):

plugin-pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
```diff
@@ -399,6 +401,17 @@ private String getRunField(RunDto run, Function<RunDto, String> getField) {
     return val.orElse("");
   }
 
+  @ShesmuMethod(
+      name = "is_clinical_pipeline",
+      description = "Check is a project pipeline would be considered clinical.")
```
avarsava, line 406: is -> if?

avarsava (approved):

PR #1049, opened 2021-04-07
Add list-to-tuple operation

apmasell:
Allow converting a list of items into a tuple of a particular length.

### DIFF
--- a/language.md
+++ b/language.md
@@ -613,6 +613,17 @@ JSON to the matching Shesmu type. Since the conversion from JSON to Shesmu
 cannot be guaranteed, it will return an optional of _type_. To create a JSON
 `null` value, use `` ` ` As json ``.
 
+- _expr_ `As {` _count_ [`,`] `}`
+Takes the values in _expr_, which must be a list, into an optional tuple with
+_count_ elements. If the list does not have enough elements, the empty optional
+is returned. If the right number is provided, they will be converted as a tuple
+in arbitrary order. If `,` is included after the target count, lists with extra
+values will be converted and the extra values discarded; otherwise, a list with
+too many values will also result in an empty optional.
+
+It is worth remembering that Shesmu removes duplicate values, so `[1, 1] As
+{2}` will return an empty optional because `[1, 1]` only has one item in it.
+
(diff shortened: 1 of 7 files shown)


language.md
```diff
@@ -613,6 +613,17 @@ JSON to the matching Shesmu type. Since the conversion from JSON to Shesmu
 cannot be guaranteed, it will return an optional of _type_. To create a JSON
 `null` value, use `` ` ` As json ``.
 
+- _expr_ `As {` _count_ [`,`] `}`
+Takes the values in _expr_, which must be a list, into an optional tuple with
+_count_ elements. If the list does not have enough elements, the empty optional
+is returned. If the right number is provided, they will be converted as a tuple
+in arbitrary order. If `,` is included after the target count, lists with extra
```
callunity, line 620: About the arbitrary order: from my reading of [Savo's comment](https://oicr.slack.com/archives/CM6TS0TLM/p1617818292026200), it sounded like order is relevant, and that's why the Min/Max solution for a list of two items was rejected. Did that turn out to not be the case?

slazicoicr: The order of a list is important for Michelle's (and I would assume most people's) use cases.  Converting an ordered list that allows duplicates to an unordered tuple with duplicates removed will cause issues when accessing via index.

apmasell: Tuples are ordered by definition, so that's not possible. Adding ordered lists would be a massive undertaking and, in her case, the input isn't ordered, so having an ordered list wouldn't help. We could change this to support a sorted stream. Something like this:

```
For x In xs:
   Sort x
   Tuple x Length 5
```
Then there would be an enforced order by the programmer. That will require the string ordering in #1050.

This would also let you take a regular expression apart into something ordered.
```
For x Splitting str By /,/:
   Tuple x Length 5
```

We can also always enforce the "must be exactly 5" because if you want to throw away extra values, you can do that explicitly using:

```
For x In xs:
   Sort x
   Limit 5
   Tuple x Length 5
```

apmasell: Yeah. Let me implement it and see if there are any gotchas.

PR #1056, opened 2021-04-08
Use new chemistry to create `flowcell_geometry` variable

apmasell:
We currently get this information out of Run Scanner, but this is awkward since
Pinery should have all this information. Now that is exported by Pinery, it can
be directly exported as a variable.

### DIFF
--- a/demo/demo.pinery_ius-input
+++ b/demo/demo.pinery_ius-input
@@ -9,6 +9,11 @@
     "dv200": null,
     "external_donor_id": "",
     "external_tissue_id": null,
+    "flowcell_geometry": [
+      [
+        1
+      ]
+    ],
     "group_desc": "",
     "group_id": "",
     "instrument_model": "Illumina MiSeq",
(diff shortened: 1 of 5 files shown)


demo/demo.pinery_ius-input
```diff
@@ -1206,6 +1353,14 @@
     "dv200": null,
     "external_donor_id": "ABC_0001",
     "external_tissue_id": null,
+    "flowcell_geometry": [
```
avarsava, line 1356: Will this ever be, say, [1, 4, 2, 3]? I'm confused why it is this format rather than an int length

apmasell: No. This format is kind of awkward, but...ughhh. So, if there are 4 compartments, then it could be that everything is joined:
`[[1, 2, 3, 4]]` or everything is split `[[1], [2], [3], [4]]`

There's a possibility that Illumina makes a half-joined flowcell, like `[[1, 2], [3, 4]]`, but I don't think that has happened.

pruzanov (approved):

PR #1073, opened 2021-04-22
Redesign thread pools

apmasell:
First, this collapses all the thread factories used by thread pools into a
single class. It then adjust the totals so that actions and olives each get
half-minus-one CPUs. Internal processes can run on all the CPUs since they
spend most of their time sleeping. Web requests can exceed the number of CPUs
since they spend most of their time blocking on I/O.

### DIFF
--- a/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/ActionProcessor.java
+++ b/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/ActionProcessor.java
@@ -288,7 +288,7 @@ public final class ActionProcessor
   }
 
   public static final int ACTION_PERFORM_THREADS =
-      Math.max(1, Runtime.getRuntime().availableProcessors() * 5 - 1);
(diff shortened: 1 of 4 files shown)


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/ActionProcessor.java
```diff
@@ -288,7 +288,7 @@ protected final boolean check(Action action, Information info) {
   }
 
   public static final int ACTION_PERFORM_THREADS =
-      Math.max(1, Runtime.getRuntime().availableProcessors() * 5 - 1);
+      Math.max(1, Runtime.getRuntime().availableProcessors() / 2 - 1);
```
pruzanov, line 291: I guess it's switching to a more frugal CPU-eating diet as discussed today?

pruzanov (approved):

callunity (approved):

asofmosq (approved):

PR #1090, opened 2021-05-25
Register workflows on migration

apmasell:
* Allow polling Vidarr during migration action
  Only do the hard work once and then allow the migration action to keep polling Vidarr.
* Automatically register workflows during migration
  Rather than pre-registering the workflows, register them once as part of migration.


apmasell:
You'll want to make the comparison white-space insensitive.

callunity (approved):

asofmosq (approved):

PR #1110, opened 2021-06-07
Expand implementation documentation

apmasell:


implementation.md
```diff
@@ -259,6 +262,13 @@ All plugin integration is provided by:
 - adding annotated methods in `PluginFile`
 - using the `Definer`
 
+A number of plugin features can be added to the `PluginFileType` or the
+`PluginFile`. Anything placed in the `PluginFileType` will be a global
+defintion. The Shesmu adminstrator can create configuration files that will
```
avarsava, line 267: This makes it sound a little like the admin is purposefully defining 'PluginFile' instances somewhere in the config files. Every configuration file automatically spawns an instance of (a child of) PluginFile, right?

implementation.md
```diff
@@ -419,8 +429,28 @@ is used to discover the properties of the action.
 
 For details on the parameters to an action, see below.
```
avarsava, line 430: This should probably refer to the section by name now, since 'below' is unrelated

implementation.md
```diff
@@ -419,8 +429,28 @@ is used to discover the properties of the action.
...
 
+These two methods are a bit different in how they operate. Actions provided by
+a method are limited by the instances, or one globally. For instance, the SFTP
```
avarsava, line 433: "Actions provided by a method are limited by the instances, or one globally."
I wish I could provide feedback other than 'this isn't parsing for me' but I can't work out what this means.

implementation.md
```diff
@@ -419,8 +429,28 @@ is used to discover the properties of the action.
...
+plugin allows creating file deletion commands. There's only one way to delete a
+file, so it is in a method per instance. The `Definer` method allows
```
avarsava, line 435: Since there's only one way to delete a file, shouldn't it be global?

implementation.md
```diff
@@ -419,8 +429,28 @@ is used to discover the properties of the action.
...
+dynamically creating actions. For instance, the Vidarr plugin will scan the
+available workflows on a remote instance and create on action definition for
```
avarsava, line 437: on -> an? on -> one?

implementation.md
```diff
@@ -419,8 +429,28 @@ is used to discover the properties of the action.
 
 For details on the parameters to an action, see below.
 
+These two methods are a bit different in how they operate. Actions provided by
+a method are limited by the instances, or one globally. For instance, the SFTP
+plugin allows creating file deletion commands. There's only one way to delete a
+file, so it is in a method per instance. The `Definer` method allows
+dynamically creating actions. For instance, the Vidarr plugin will scan the
```
avarsava, line 436: _The_ Vidarr plugin? Is this meant to say server or is this a particular plugin?

implementation.md
```diff
@@ -419,8 +429,28 @@ is used to discover the properties of the action.
...
+available workflows on a remote instance and create on action definition for
+each workflow type. The `Definer` allows creating custom parameters, so in the
+workflow case, some parameters are baked into the Vidarr submit action, but
```
avarsava, line 439: So at this point some piece of code has scanned a remote instance and created Actions for each workflow type. In this context I am not sure what 'in the workflow case' means, because it sounds like we only have workflows to work with

language.md
```diff
@@ -397,10 +397,6 @@ Check if _expr_ is true for all rows. If none are collected, the result is true.
 
 Count the number of matched rows
 
-- `First` _expr_
```
avarsava, line 400: There's usages of First in tutorial.md

avarsava (approved):

implementation.md
apmasell: Unmatched `}` on line 433.

I've tried to rewrite this whole block.

avarsava (approved):
Looks good!

PR #1116, opened 2021-06-10
Prioritize QUEUED and INFLIGHT actions as they need to move quickly

avarsava:
Once they've been sorted by time-since-last-check, further bucket by lastState - QUEUED and INFLIGHT get greater priority in order to reflect cromwell status better and set off downstream workflows faster. Previously the scheduler had a chance of only looking at UNKNOWN actions and never checking the real work getting done.

Java promises sorted() otherwise preserves internal order but idk if i trust it. I'm not sure if this undoes the previous sort instead of preserving it inside the two tiers i'm trying to build. 

This change is untested! I don't have a local environment with actual actions going through

### DIFF
--- a/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/ActionProcessor.java
+++ b/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/ActionProcessor.java
@@ -1642,6 +1642,24 @@ public final class ActionProcessor
                                     .getSeconds()
                                 / 600)
                     .thenComparingInt(e -> e.getKey().priority()))
+            .sorted(
+                new Comparator<>() {
+                  @Override
+                  public int compare(Entry<Action, Information> o1, Entry<Action, Information> o2) {
+                    ActionState o1State = o1.getValue().lastState,
+                        o2State = o2.getValue().lastState;
+                    // Consider INFLIGHT and QUEUED to be Equal but greater than the rest
+                    if (o1State == o2State
+                        || (o1State == ActionState.INFLIGHT && o2State == ActionState.QUEUED)
+                        || (o1State == ActionState.QUEUED && o2State == ActionState.INFLIGHT))
+                      return 0;
+                    // At this point we know either o1State or o2State is one of the other ones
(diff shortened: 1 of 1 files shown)


djcooke (approved):
I haven't dug through the Shesmu code to figure out what `ActionState#sortPriority` is used for, but it would be ideal to either change those values and use that, or add a "processPriority" field to `ActionState` and use that instead via `stream.sort(Comparator.comparingInt(x -> x.getValue().lastState.processPriority()))`. This would help to keep the `update` method closer to a reasonable length and also simplify future tweaking.

If you'd rather not, it's fine as is.

> Java promises sorted() otherwise preserves internal order but idk if i trust it. I'm not sure if this undoes the previous sort instead of preserving it inside the two tiers i'm trying to build.

Java is good at sticking to its contracts. The first sort is applied, so you have a sorted stream getting passed to the second sort. Any items that return 0 when compared are left in their original order.

avarsava:
> I haven't dug through the Shesmu code to figure out what `ActionState#sortPriority` is used for, but it would be ideal to either change those values and use that, or add a "processPriority" field to `ActionState` and use that instead via `stream.sort(Comparator.comparingInt(x -> x.getValue().lastState.processPriority()))`. This would help to keep the `update` method closer to a reasonable length and also simplify future tweaking.
> 
> If you'd rather not, it's fine as is.

sortPriority seems like it's for organizing the UI? As far as I can tell it's applied in scheduling priority at all. I'll make that change!

> Java is good at sticking to its contracts. The first sort is applied, so you have a sorted stream getting passed to the second sort. Any items that return 0 when compared are left in their original order.

OK sounds good, thanks!

callunity (approved):

avarsava:
> I haven't dug through the Shesmu code to figure out what `ActionState#sortPriority` is used for, but it would be ideal to either change those values and use that, or add a "processPriority" field to `ActionState` and use that instead via `stream.sort(Comparator.comparingInt(x -> x.getValue().lastState.processPriority()))`. This would help to keep the `update` method closer to a reasonable length and also simplify future tweaking.

Updated

djcooke (approved):

PR #1120, opened 2021-07-09
Check for null WorkflowRuns because of Niassa fetch fails

morgantaschuk:


plugin-niassa/src/main/java/ca/on/oicr/gsi/shesmu/niassa/NiassaServer.java
```diff
@@ -165,6 +165,9 @@ public DirectoryAndIniCache(Path fileName) {
       final MetadataWS metadata = metadataConstructor.get();
       final WorkflowRun run = metadata.getWorkflowRun(key.intValue());
       metadata.clean_up();
+      if (run == null) {
+        return null;
```
callunity, line 169: Looking further down the method, there are multiple `return Optional.empty()` , even in the case of catching an exception (line 202). Returning `Optional.empty()` here might be more idiomatic, and calling code already knows to check for empty optional.

plugin-niassa/src/main/java/ca/on/oicr/gsi/shesmu/niassa/NiassaServer.java
```diff
@@ -628,7 +631,15 @@ public void configuration(SectionRenderer renderer) {
   }
 
   public WorkflowRunEssentials directoryAndIni(long workflowRun) {
-    return directoryAndIniCache.get(workflowRun).orElse(WorkflowRunEssentials.EMPTY);
+    WorkflowRunEssentials wre =
+        directoryAndIniCache.get(workflowRun).orElse(WorkflowRunEssentials.EMPTY);
+    if (wre == WorkflowRunEssentials.EMPTY) {
+      definer.log(
+          String.format(
+              "NiassaServer.DirectoryAndIniCache.fetch: Niassa Workflow Run fetch failed for WorkflowRun SWID: %d"),
```
callunity, line 639: Is the String.format call intended to have the `workflowRun` accession as its second argument?

callunity (approved):

djcooke (approved):

PR #1122, opened 2021-08-05
fixed cases where HttpExchange was not closed

djcooke:
`HttpExchange` javadoc:

> Exchanges are terminated when both the request InputStream and response OutputStream are closed.
> Closing the OutputStream, implicitly closes the InputStream (if it is not already closed).
> ...
> The effect of failing to terminate an exchange is undefined, but will typically result in resources failing to be freed/reused.

### DIFF
--- a/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
+++ b/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
@@ -61,7 +61,6 @@ import java.net.*;
 import java.net.http.HttpClient;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
-import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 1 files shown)


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
```diff
@@ -2965,27 +2965,6 @@ public ObjectNode savedSearches() {
     return searchInfo;
   }
 
-  private void showSourceConfig(XMLStreamWriter writer, Path filename) {
```
avarsava, line 2968: Removing this whole method?

avarsava (approved):

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
djcooke: It was private and unused

callunity (approved):

PR #1124, opened 2021-08-10
Add pinery_ius_all input format to be able to operate on skipped samples and lanes

callunity:
GP-2807

The pinery_ius_all input format is almost exactly the same as the pinery_ius input format, except that it doesn't initially filter out skipped samples, and it also exposes a "skip" attribute that can be used to filter in olives.

At this point we expect that analysis olives will continue to use the pinery_ius input format, and only a small set of olives (e.g. Dashi) will use this new pinery_ius_all input format.

I'm open to naming suggestions for this input format!


asofmosq (approved):

plugin-pinery/README.md
```diff
@@ -2,9 +2,12 @@
 [Pinery](http://github.com/oicr-gsi/pinery) is a web service application that
 provides generalized LIMS (Laboratory Information Management System) access for information about samples.
 
-The Pinery plugin provides two input formats:
+The Pinery plugin provides three input formats:
 
-- `pinery_ius` contains lane and sequenced sample information
+- `pinery_ius` contains lane and sequenced sample information, and excludes 
+   skipped lanes and samples
+- `pinery_ius_all` contains lane and sequenced sample information, including
```
djcooke, line 9: renamed

plugin-pinery/README.md
```diff
@@ -18,9 +21,9 @@ To configure a Pinery source, create a JSON file ending in `.pinery` as follows:
     }
 
 where `provider` is an arbitrary string that will be baked into the
-`pinery_ius`'s `provider` field. `shortProvider` is similar, but will be used
-in `external_key`. `url` is the address of the Pinery server and `version`
-provides the Pinery data model version.
+`provider` field for `pinery_ius` and `pinery_ius_all`. `shortProvider` is
```
djcooke, line 24: `pinery_ius_all` renamed

djcooke (approved):

plugin-pinery/README.md
callunity: fixed this and the other occurrence in the file

PR #1138, opened 2021-11-16
[GP-2862] Unload & Purge on SUCCESSFUL Vidarr actions

avarsava:


avarsava:
I can't figure out how to switch a PR back to a draft but I need to check recursive purging

avarsava:
TODO:

- Separate PR on vidarr for returning list of unloaded workflow-run ids in unload endpoint (add RETURNING clause to db query in Main.unload(), output in endpoint)
- finish the murder call tagged with FIXME in SubmitAction.java: should be something like `builder.and(Stream.of(builder.type(Stream.of("vidarr-run")), builder.tags(response.getDeletedWorkflowRunIds().stream().map("vidarr-workflow-run"::concat)))` (this particular formatting should allow for matching against RunStateMonitor's vidarr-workflow-id tag
- Also (probably in another PR) implement a generic Murder button (_please_ call it something else in the front end) to check whether a vidarr workflow run exists and delete the shesmu action if it does not. You should probably rename 'Search Vidarr Again' as that checks whether a vidarr workflow run exists and _reruns_ if it does not, and these two might get confusing. The generic button is needed because unloading from one vidarr instance can affect multiple shesmu instances and the other instances will need a way to check whether workflow run ids are still relevant.
- Please update the Shesmu Ops Guide and any relevant vidarr docs as well!!

PR #1139, opened 2021-11-17
Add cerberus_fp_skipped input format

asofmosq:


plugin-niassa/src/main/java/ca/on/oicr/gsi/shesmu/pipedev/BaseProvenancePluginType.java
```diff
@@ -206,15 +207,192 @@ public ItemCache() {
...
+        final AtomicInteger badFilePaths = new AtomicInteger();
+        final AtomicInteger badSets = new AtomicInteger();
+        final AtomicInteger badVersions = new AtomicInteger();
+        final Map<String, Integer> badSetCounts = new TreeMap<>();
+        return client
+            .map(BaseProvenancePluginType.this::fetch)
+            .orElseGet(Stream::empty)
+            .filter(
+                fp ->
+                    (fp.getSkip() == null || fp.getSkip().equals("true"))
```
callunity, line 229: It looks like the standard CerberusFPValue also selects items where `fp.getSkip() == null`. Should those records be included in both CFPValue and CFPSkippedValue caches?

plugin-cerberus/src/main/java/ca/on/oicr/gsi/shesmu/cerberus/BaseCerberusFileProvenanceSkippedRecord.java
```diff
@@ -0,0 +1,149 @@
...
+    final var workflowRunLabels = new TreeMap<String, JsonNode>();
+    final var labels = workflow.getLabels().fields();
+    while (labels.hasNext()) {
+      final var label = labels.next();
+      workflowRunLabels.put(label.getKey(), label.getValue());
+    }
+    return workflowRunLabels;
+  }
+
+  // Do we care about stale records if they've been failed?
```
callunity, line 32: My guess is no, but I hope @mlaszloffy can speak to this

plugin-niassa/src/main/java/ca/on/oicr/gsi/shesmu/pipedev/BaseProvenancePluginType.java
asofmosq: They shouldn't get included in CFPSkippedValue as those records won't correspond with actually failed samples. I'll push an amendment.

callunity (approved):

plugin-cerberus/src/main/java/ca/on/oicr/gsi/shesmu/cerberus/BaseCerberusFileProvenanceSkippedRecord.java
mlaszloffy: I think `stale` should be included so that the input format is drop in compatible with cerberus_fp and so that the records we're trying to fetch from cerberus_fp_skipped can be inspected to see if they have gone stale or not.

plugin-cerberus/src/main/java/ca/on/oicr/gsi/shesmu/cerberus/BaseCerberusFileProvenanceSkippedRecord.java
```diff
@@ -0,0 +1,149 @@
...
+import com.fasterxml.jackson.databind.JsonNode;
+import java.nio.file.Path;
+import java.time.Instant;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+import java.util.TreeSet;
+import java.util.stream.Collectors;
+
+abstract class BaseCerberusFileProvenanceSkippedRecord<T extends LimsProvenance>
```
callunity, line 19: Could you add a `skip` field to the SkippedRecord? We had talked about joining cerberus_fp to cerberus_fp_skipped in some use cases, and it would no longer possible to tell if a record is skipped.

mlaszloffy (approved):

PR #1140, opened 2021-11-19
Allow action commands to remove other actions

apmasell:
* Allow action commands to remove other actions
  This permits a command to return a filter that triggers purging of other actions.
* Remove JSON Schema support
  This was only added for use with SONG.


avarsava (approved):

mlaszloffy (approved):
👍

PR #1156, opened 2022-03-22
Fix bug where dumpers are not cleaned up

apmasell:
There was incorrect code in `BaseOliveBuilder` for cleaning up dumpers that
tried to use fields. Because dumpers can be connected to `Define` olives, the
dumper can be invoked from shared context and should not change instance
fields. To solve this problem, this creates `Stream.onClose` handlders that
clean up any created dumpers. To accomplish this:

1. Remove all the dumper logic from `RootBuilder`.
2. Create a lambda builder for `Runnable`.
3. Add an `onClose` code generator in `BaseOliveBuilder`.
4. Expand `RejectNode` to have code generation for both the work and the
   cleanup (`onClose`) methods.
5. Extend `OliveClauseNodeBaseDump` to call `Dumper.stop` in the `onClose`
   handler.
6. Modify the testing framework to ensure that dumper cleanup is performed.


avarsava (approved):

mlaszloffy (approved):
Thanks a lot Andre! The jsondumper api is working great with your PR.

PR #1160, opened 2022-05-17
Gp 3210 use aggregate skip status from cerberus

mlaszloffy:
This won't build correctly until the cerberus 0.2.10 release.

JIRA Ticket: GP-3210

- [X] Updates Changelog
- [] Updates developer documentation

### DIFF
--- a/plugin-cerberus/pom.xml
+++ b/plugin-cerberus/pom.xml
@@ -48,7 +48,7 @@
     <dependency>
       <groupId>ca.on.oicr.gsi.cerberus</groupId>
       <artifactId>cerberus-core</artifactId>
-      <version>0.2.9</version>
+      <version>0.2.10</version>
     </dependency>
   </dependencies>
   <build>
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,7 +1,10 @@
 # Unreleased
 
+* Fix bug where dumpers are not cleaned up
 * Doc improvement to clarify how ties are handled in `Pick Max`/`Pick Min`
 * Use the aggregate skip status from cerberus to categorize cerberus_fp records
+* Runscanner -> 1.13.2
+* Vidarr -> 0.7.0
 * Cerberus -> 0.2.10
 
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -3,16 +3,16 @@ FROM debian:bullseye
...
-RUN cd /build && \
+RUN --mount=type=cache,target=/root/.m2 cd /build && \
     tsc -p shesmu-server-ui && \
-    mvn -q -Dsurefire.useFile=false -DskipIT=true clean install && \
+    mvn -Dsurefire.useFile=false -DskipIT=true clean install && \
     mkdir /usr/share/shesmu && \
     cp shesmu-pluginapi/target/shesmu-pluginapi.jar shesmu-server/target/shesmu.jar plugin-*/target/shesmu-plugin-*.jar /usr/share/shesmu
 
-FROM openjdk:16-slim
+FROM openjdk:17
```
avarsava, line 15: I have yet to verify that shesmu runs correctly with java 17?

callunity: We don't use the Dockerfile in our local deploy - I believe it's only used to build a Docker image during the release (and that hasn't been working reliably recently).

callunity (approved):

Dockerfile
avarsava: Why update it in this PR?

plugin-cerberus/pom.xml
```diff
@@ -48,7 +48,7 @@
     <dependency>
       <groupId>ca.on.oicr.gsi.cerberus</groupId>
       <artifactId>cerberus-core</artifactId>
-      <version>0.2.9</version>
+      <version>0.2.10-SNAPSHOT</version>
```
callunity, line 51: Cerberus 0.2.10 is now released, so this can be updated

Dockerfile
mlaszloffy: I was testing with the dockerfile locally, I can split this out into another PR for later.

avarsava (approved):

PR #1172, opened 2022-08-19
[GP-3379] Add gauges to track last start and end of cache refresh

avarsava:
JIRA Ticket: GP-3379

These metrics have been served on shesmu-dev for about a day, they are available in prometheus and grafana, please let me know if these are useful

- [X] Updates Changelog

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* Adds metrics for cache refresh start and end times
+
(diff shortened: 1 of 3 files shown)


avarsava:
This change permits calculations for alerts such as 'IF start time is greater than end time [ie current refresh hasn't ended] AND start time is 1H behind now() THEN alert'

callunity (approved):
Looks like they'll be really useful!

mlaszloffy (approved):

PR #1182, opened 2022-11-01
[GP-3326] other dependencies

avarsava:
JIRA Ticket: GP-3326

- [N/A] Updates Changelog

### DIFF
--- a/plugin-mongo/pom.xml
+++ b/plugin-mongo/pom.xml
@@ -22,7 +22,7 @@
     <dependency>
         <groupId>org.mongodb</groupId>
         <artifactId>mongodb-driver-sync</artifactId>
-        <version>3.11.0</version>
+        <version>3.11.3</version>
     </dependency>
     <dependency>
       <groupId>com.fasterxml.jackson.core</groupId>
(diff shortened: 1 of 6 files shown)


plugin-mongo/pom.xml
```diff
@@ -22,7 +22,7 @@
     <dependency>
         <groupId>org.mongodb</groupId>
         <artifactId>mongodb-driver-sync</artifactId>
-        <version>3.11.0</version>
+        <version>4.8.0-beta0</version>
```
callunity, line 25: Is there a more stable version available?

plugin-sftp/pom.xml
```diff
@@ -22,7 +22,7 @@
     <dependency>
       <groupId>com.hierynomus</groupId>
       <artifactId>sshj</artifactId>
-      <version>0.27.0</version>
+      <version>0.34.0</version>
```
callunity, line 25: :tada:

plugin-mongo/pom.xml
avarsava: Both surprised and not surprised that the maven dependency updater thinks beta versions are acceptable. Thanks for pointing this out

avarsava: Fixed

callunity (approved):

mlaszloffy (approved):

PR #1187, opened 2022-12-02
[GP-3619] Lock down tsc version in README

avarsava:
JIRA Ticket: GP-3619

- [n/a] Updates Changelog
- [n/a] Updates developer documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,9 @@ Build dependencies:
 
 - Java 14 or later
 - Maven 3.5 or later
-- TypeScript 3.9 or later (the version supplied with Ubuntu and Debian in
+- TypeScript 4.3.2 (the version supplied with Ubuntu and Debian in
   `node-typescript` is too old. Use `sudo npm -g i typescript` to install a
-  newer copy)
+  newer copy. However, versions newer than 4.3.2 will fail to compile.)
 - Docker (optional) for container builds
 
 Optional runtime dependencies:


callunity (approved):

djcooke (approved):
Installing this as a dev dependency in the project rather than recommending global install could both enforce the version used and simplify dev setup

avarsava:
> Installing this as a dev dependency in the project rather than recommending global install could both enforce the version used and simplify dev setup

I don't know if typescript has something like python's virtual environments, how would i do this?

djcooke:
> > Installing this as a dev dependency in the project rather than recommending global install could both enforce the version used and simplify dev setup
> 
> How would i do this?

It looks like Shesmu doesn't really use npm right now, so it would require some setup. You'd need a `package.json` with typescript as a dev dependency. You could then run tsc via npx.

You could also take it a step farther and use exec-maven-plugin to do your ts build automatically as part of the Maven build.

avarsava:
> It looks like Shesmu doesn't really use npm right now, so it would require some setup. You'd need a `package.json` with typescript as a dev dependency. You could then run tsc via npx.
> 
> You could also take it a step farther and use exec-maven-plugin to do your ts build automatically as part of the Maven build.

We have a ticket (GP-3477) for using exec-maven-plugin but it's marked as requiring tsc 4.4, which in turn needs development work because the front end doesn't compile with that version. 
This will all be resolved when we get around to that ticket, sound OK?

PR #1188, opened 2022-12-07
Fix using signatures after `Flatten` and `Require`

apmasell:
Changes incorrect field types for dynamic signatures that causes a missing field when the correct `GETFIELD` bytecode is executed.

JIRA Ticket: 

- [X] Updates Changelog

### DIFF
--- a/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/FlattenBuilder.java
+++ b/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/FlattenBuilder.java
@@ -1,5 +1,7 @@
 package ca.on.oicr.gsi.shesmu.compiler;
 
+import static ca.on.oicr.gsi.shesmu.compiler.TypeUtils.TO_ASM;
+
 import ca.on.oicr.gsi.shesmu.compiler.definitions.InputVariable;
 import ca.on.oicr.gsi.shesmu.compiler.definitions.SignatureDefinition;
 import java.util.function.Consumer;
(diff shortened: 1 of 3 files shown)


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/FlattenBuilder.java
```diff
@@ -51,7 +53,7 @@ public FlattenBuilder(
             Stream.concat(
                     Stream.of(originalType, unrollType),
                     copySignatures
-                        ? owner.signatureVariables().map(SignatureDefinition::storageType)
+                        ? owner.signatureVariables().map(s -> s.type().apply(TO_ASM))
```
avarsava, line 56: Could you please explain more why this fix works? Why was 'storageType' incorrect and why does converting the type to ASM fix it?

avarsava:
PR title says 'after Flatten and Require' but the only code change is in `FlattenBuilder`. Do both Flatten and Require get built using `FlattenBuilder`?

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/FlattenBuilder.java
apmasell: Signatures kind of function in different modes. When you access a `std::signature::sha1`, it's a `string` in Shesmu or `java.lang.String` in Java. However, there's a bunch of places where Shesmu needs to process signatures in Java code over all the incoming records, so, we don't pass a string, but a function that returns a string. That isn't true for all signatures though. Some signatures are row-independent (static signatures), in which case we _do_ get the actual type. `storageType()` knows all this and figures it out.

Unfortunately, in this case, we are generating output rows, so we want to get the real type because the calling code is going to generate the signature for each row and it needs to be copied into the new row type we are building. The exception was that the field was missing; this is only partially true. The field existed, but it had the wrong type (and name+type combinations have to be unique in the JVM, not only name as Java requires).

So `type()` gets the real Shesmu type for this signature and then we convert it to a JVM type that the ASM library can write out.

apmasell:
Yes. `Require` is implemented a special kind of `Flatten` that iterates over a `java.util.Optional` rather than a `java.util.Set`. `FlattenBuilder` consumes a `Stream`, in fact, and the two different pieces of syntax generate streams appropriately.

avarsava (approved):

mlaszloffy (approved):

PR #1203, opened 2023-02-02
Add accredited_projects and accredited_with_clinical_report_projects functions

callunity:
JIRA Ticket: GP-3731

- [x] Updates Changelog

`N/A` Updates developer documentation

This PR is in service of being able to write an alert that only considers Accredited with Clinical Report projects.

### DIFF
--- a/plugin-pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
+++ b/plugin-pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
@@ -734,12 +734,39 @@ public class PinerySource extends JsonPluginFile<PineryConfiguration> {
 
   @ShesmuMethod(
       name = "clinical_projects",
-      description = "Projects marked clinical from in Pinery defined in {file}.")
+      description = "Projects marked clinical from Pinery defined in {file}.")
   public Set<String> clinicalProjects() {
     return projects
         .get()
         .filter(
             project -> project.getPipeline() != null && isClinicalPipeline(project.getPipeline()))
+        .map(SampleProjectDto::getName)
+        .collect(Collectors.toCollection(TreeSet::new));
+  }
+
+  @ShesmuMethod(
+      name = "accredited_projects",
+      description = "Projects with Accredited pipeline from Pinery defined in {file}.")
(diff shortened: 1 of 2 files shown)


mlaszloffy (approved):

plugin-pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
```diff
@@ -744,6 +744,32 @@ public Set<String> clinicalProjects() {
...
+        .get()
+        .filter(
+            project -> project.getPipeline() != null && "Accredited".equals(project.getPipeline()))
+        .map(SampleProjectDto::getName)
+        .collect(Collectors.toCollection(TreeSet::new));
+  }
+
+  @ShesmuMethod(
+      name = "accredited_with_clinical_report_projects",
+      description = "Projects with Accredited pipeline from in Pinery defined in {file}.")
```
avarsava, line 761: Description could reflect 'with Report'

avarsava (commented):
The 'from in Pinery' construction reads weird but it's consistent.
isn't 'Accredited with Report' replacing 'Clinical'?

callunity:
> The 'from in Pinery' construction reads weird but it's consistent. isn't 'Accredited with Report' replacing 'Clinical'?

Ah yep, I'll update all those descriptions. Definitely copypasta'd that.
"Accredited with Clinical Report" and "Accredited" are both clinical pipelines, but only AwCR samples are "true" clinical samples.

plugin-pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
callunity: done

avarsava:
> > The 'from in Pinery' construction reads weird but it's consistent. isn't 'Accredited with Report' replacing 'Clinical'?
> 
> Ah yep, I'll update all those descriptions. Definitely copypasta'd that. "Accredited with Clinical Report" and "Accredited" are both clinical pipelines, but only AwCR samples are "true" clinical samples.

Right, so just want to double check, does the `clinical_projects` method still make sense?

avarsava (approved):

PR #1210, opened 2023-03-17
[GP-3706] Carry plugin error to server

avarsava:
JIRA Ticket: GP-3706

**I think close to done**, just want eyes on this.

We need a way to communicate error status between the plugins and the server. The PluginManager is the class that's responsible for handling communications, and it does this almost entirely with Streams. So, create a new Stream implementation that also carries around a boolean that can either be set to false or becomes false if it comes in contact with another bad stream. 
Plugins may choose what conditions are 'not ok'. In the case of Cerberus, use the BaseRecord's 'invalidated' state, which is the 'lastUpdated' timestamp being equal to Instant.EPOCH. I will update the other plugins too; I'm not done.
The PluginManager doesn't need to change since it's all Stream to it.
The server will then decide what to do if it encounters a stream that's marked 'not ok'. I need to refactor AnnotatedFormatDefinition's 'writeJson', the input endpoint in the Server is the only thing that calls it and the order of operations is going to need to change in order to get the header code set correctly. 

also i have so much documenting to do
also deletes some orphaned classes
also fixes some typos

- [ ] Updates Changelog


avarsava:
Unsafe to merge until we decide what we're going to do about code formatting - several classes here break the googlecodeformat git hook because they use newer language features

avarsava:
Still unsafe to merge because of the aforementioned formatting problems 

BUT now passes the following tests:
- /input and input/stale return 503 when InitialCachePopulationException for remote endpoints
- returns 200, stale data when remote endpoint breaks after initial population (ie, cache fallback still works)
- /input and input/stale return 503 when InitialCachePopulationException for cerberus plugin
- returns 200, stale data when cerberus plugin breaks after initial population (ie, cache fallback still works)
- when federating local and remote cerberus inputs, either one being broken returns 503 on input and input/stale
- when federating local and remote cerberus inputs, when both are working, 200 and data
- one input format (ie cerberus) returning 503 does not interfere with another (ie pinery) returning 200

Lmk if there's any cases you think I missed but I'm pretty confident this will work as intended.

Outstanding question about semaphores in the PR
Still should document some of these systems
BUT i think this is ready for review again

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
```diff
@@ -2631,15 +2633,33 @@ public void downloadInputData(
...
+    if (fetchedInput.isOk()) {
+      t.getResponseHeaders().set("Content-type", "application/json");
+      t.sendResponseHeaders(200, 0);
+    } else {
+      t.sendResponseHeaders(503, 0);
+      try (var os = t.getResponseBody()) {}
+        /**
+         * FIXME
+         * Previously, writeJson included a call to fetch, now we download the input much earlier.
+         * Do i need to release the inputDownloadSemaphore for any failures here too?
```
callunity, line 2650: By "failures here" do you mean failure to get the response body, and releasing the inputDownloadSemaphore in a corresponding catch?

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/plugins/AnnotatedInputFormatDefinition.java
```diff
@@ -308,13 +309,14 @@ public RemoteReloader(Path fileName) {
 
       @Override
       protected Stream<Object> fetch(Instant lastUpdated) throws Exception {
-        if (config.isEmpty()) return Stream.empty();
+        if (config.isEmpty()) return Stream.empty(); // probably not OK
```
callunity, line 312: Agreed, probably not ok

plugin-guanyin/src/main/java/ca/on/oicr/gsi/shesmu/guanyin/GuanyinRemote.java
```diff
@@ -31,6 +31,7 @@ public ReportsCache(Path fileName) {
 
     @Override
     protected Stream<GuanyinReportValue> fetch(Instant lastUpdated) throws Exception {
+      // TODO: any uses for ErrorableStream here?
```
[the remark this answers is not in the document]
avarsava: I think you're the guanyin expert, so if you think it would make sense, we can do that

callunity:
And re. formatting problems - it looks like we're using `maven-git-code-format` v1.39. The project looks to have been renamed to `hgit-code-format-maven-plugin` and is now on v4.2, so it would be worth looking into updating the plugin.

plugin-guanyin/src/main/java/ca/on/oicr/gsi/shesmu/guanyin/GuanyinRemote.java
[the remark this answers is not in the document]
callunity: Sounds good, I think that would give consistency between plugins

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
avarsava: by 'failures here' i mean this case where !fetchedInput.isOk and we're returning 503

avarsava: Currently I have it releasing the semaphore but in the debugger it looks like that makes the semaphore value climb beyond what it was initially set to and i'm not sure that's supposed to happen

avarsava:
> git-code-format-maven-plugin

This looks promising, but it appears that it requires some per-user setup if you want to use JDK 17: https://github.com/Cosium/git-code-format-maven-plugin#jdk-16-peculiarities

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
callunity: That debugger output is indeed weird. The `return` on line 2654 means the second `release` on line 2662 shouldn't be tripped, and I can't see how the `release` on line 2653 would be called twice. Is this a debugger running on a test, or have you deployed locally and attached the debugger to check out a more real-world query?

avarsava: I observed this with shesmu deployed locally and using the /input and /input/stale endpoints, which are what will be used in practice as well

avarsava:
Formatting concerns resolved

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/plugins/AnnotatedInputFormatDefinition.java
```diff
@@ -308,13 +309,15 @@ public RemoteReloader(Path fileName) {
...
       protected Stream<Object> fetch(Instant lastUpdated) throws Exception {
-        if (config.isEmpty()) return Stream.empty();
+        if (config.isEmpty()) return Stream.empty(); // FIXME probably not OK
         final var url = config.get().getUrl();
         var response =
             Server.HTTP_CLIENT.send(
                 HttpRequest.newBuilder(URI.create(url)).GET().version(Version.HTTP_1_1).build(),
                 BodyHandlers.ofInputStream());
         try (var parser = RuntimeSupport.MAPPER.getFactory().createParser(response.body())) {
+          // TODO do i need to test that it's 2xx rather than specifically 200?
```
callunity, line 319: Looking at the other 200-series codes, only 200 OK is guaranteed to return usable data.

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
callunity: That does indeed sound like a weird bug, the semaphore counter shouldn't go beyond what it was initially set to.

callunity: If you don't release the semaphore here, does the semaphore eventually run out of available slots?

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
```diff
@@ -2631,11 +2633,28 @@ public void downloadInputData(
...
+      t.sendResponseHeaders(200, 0);
+    } else {
+      t.sendResponseHeaders(503, 0);
+      try (var os = t.getResponseBody()) {}
+      /**
+       * FIXME Previously, writeJson included a call to fetch, now we download the input much
+       * earlier. Do i need to release the inputDownloadSemaphore for any failures here too? I think
+       * this increments the semaphore too much but I can't tell what the proper behaviour is
+       */
+      inputDownloadSemaphore.release();
```
djcooke, line 2652: * `os` is unused
* semaphore acquisition and release looks like it should be fine to me. If a permit isn't acquired, the method returns early. If a permit is acquired, it should be released in one of two places. I don't see how it could be released twice

avarsava: Looks like getting the response body closes the stream (https://docs.oracle.com/en/java/javase/17/docs/api/jdk.httpserver/com/sun/net/httpserver/HttpExchange.html#getResponseBody()) and in its original implementation it nicely echoed the format of the else block: https://github.com/oicr-gsi/shesmu/pull/519/files#diff-8800ecd9470735edffb75cb231ef63395c02e230a1e4070f5fc9a0e90217c3d8

Could see about calling close() instead, but it looks like this is a valid way to close the HttpExchange too

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/ErrorableStream.java
```diff
@@ -0,0 +1,306 @@
...
+  public ErrorableStream(Stream<T> inner) {
+    this.inner = inner;
+    if (inner instanceof ErrorableStream<T>) {
+      this.ok = ((ErrorableStream<T>) inner).ok;
+    } else {
+      this.ok = true;
+    }
+  }
+
+  public void invalidate() {
```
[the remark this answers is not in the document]
avarsava: Could be! Each plugin behaves differently. I'd rather offer options in the API early rather than have a plugin author need to change the API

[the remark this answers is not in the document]
djcooke: Plugins can do whatever they want internally. Will Shesmu ever use data from an invalid `ErrorableStream`?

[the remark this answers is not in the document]
avarsava: Oh, that's true, the server returns early and doesn't write out what json it can. I could change that

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
djcooke: Ah, ok. Getting the response body doesn't close the stream, but it's in a try-with-resources statement, so it gets auto-closed after that (empty) block completes. Calling `close` would have a more obvious purpose

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
avarsava: Looks like it. 
Testing more, this happens specifically when readStale is true because it never hits the tryAcquire at 2631.

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/ErrorableStream.java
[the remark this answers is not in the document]
djcooke: Would you want to change that? The data is marked as bad, and there's no indication as to why. I would assume I don't want to use that data for anything

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
callunity: Ok, I'll keep an eye out for a commit that fixes this.

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
```diff
@@ -2649,7 +2649,7 @@ public void downloadInputData(
        * earlier. Do i need to release the inputDownloadSemaphore for any failures here too? I think
        * this increments the semaphore too much but I can't tell what the proper behaviour is
        */
-      inputDownloadSemaphore.release();
+      if (!readStale) inputDownloadSemaphore.release();
```
callunity, line 2652: Great! Would you mind removing the FIXME comment as well?

avarsava: fixed

callunity (approved):
Major kudos to you on working through this tricky PR! It looks great :tada:

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/BaseSimulateRequest.java
```diff
@@ -542,9 +543,12 @@ public void function(
                                 Function.identity(),
                                 name -> {
                                   CACHE_REFRESH.labels(readStale ? "cached" : "fresh", name).inc();
-                                  return inputSource
-                                      .fetch(name, readStale)
-                                      .collect(Collectors.toList());
+                                  final var results = inputSource.fetch(name, readStale);
+                                  if (results instanceof ErrorableStream<Object> errorableResults) {
+                                    if (!errorableResults.isOk())
+                                      throw new RuntimeException("Server error fetching data");
```
callunity, line 549: Would it be possible to include the name of which input source errored? (In case there's a systematic error that needs to be tracked down)

avarsava: yes, good idea, fixed

djcooke (approved):
I think checked exceptions would be much simpler and have several advantages, but I understand Shesmu's architecture doesn't support them.

This looks fine to me where you've chosen to create and check for the ErrorableStreams, though there is no way to know whether they've been overlooked in other places

PR #1211, opened 2023-04-01
Redesign input cache population

apmasell:
Change the way that the required input formats are extracted and placed in cache for the olives. The goal of this is to prevent an olive from running when its input was not cached.

- [X] Updates Changelog


avarsava (approved):

mlaszloffy (approved):
Thanks Andre!

PR #1222, opened 2023-07-14
Add case_archive input format

callunity:
JIRA Ticket: GP-3994

- [x] Updates Changelog
- [x] Updates developer documentation

I'm not super attached to the `Nabu` prefix for `NabuCaseArchiveFormatDefinition`, and could certainly remove that if you think that would be clearer.

Can test this out against nabu-dev by creating a `infrastructure/shesmu/local/local.case_archive-remote` file with:
```
  { "url": "https://nabu-dev.gsi.oicr.on.ca/cases", "ttl": 30 } 
```

### DIFF
--- a/plugin-nabu/src/main/java/ca/on/oicr/gsi/shesmu/nabu/NabuCaseArchiveValue.java
+++ b/plugin-nabu/src/main/java/ca/on/oicr/gsi/shesmu/nabu/NabuCaseArchiveValue.java
@@ -1,0 +1,101 @@
+package ca.on.oicr.gsi.shesmu.nabu;
+
+import ca.on.oicr.gsi.shesmu.plugin.input.ShesmuVariable;
+import java.time.Instant;
+import java.util.Optional;
+import java.util.Set;
+
+public class NabuCaseArchiveValue {
+  private final Optional<Instant> caseFilesUnloaded;
+  private final String caseIdentifier;
+  private final Optional<String> commvaultBackupJobId;
+  private final Instant created;
+  private final Optional<Instant> filesCopiedToOffsiteArchiveStagingDir;
+  private final Optional<Instant> filesLoadedIntoVidarrArchival;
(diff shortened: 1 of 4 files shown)


plugin-nabu/src/main/java/ca/on/oicr/gsi/shesmu/nabu/NabuCaseArchiveValue.java
```diff
@@ -0,0 +1,101 @@
...
+      Optional<String> commvaultBackupJobId,
+      Instant created,
+      Optional<Instant> filesCopiedToOffsiteArchiveStagingDir,
+      Optional<Instant> filesLoadedIntoVidarrArchival,
+      Set<String> limsIds,
+      Instant modified,
+      long requisitionId,
+      Set<String> workflowRunIdsForOffsiteArchive,
+      Optional<Set<String>> workflowRunIdsForVidarrArchival) {
+    super();
```
avarsava, line 33: What's the role of super() here? in my IDE it opens Object.class

avarsava (approved):

plugin-nabu/src/main/java/ca/on/oicr/gsi/shesmu/nabu/NabuCaseArchiveValue.java
callunity: I'm not exactly sure; this is following the same pattern used by other `*Value` classes (e.g. [PineryIUSValue](https://github.com/oicr-gsi/shesmu/blob/d2412f304ffb88b5884aa5c500fdbe4e2b003ea0/plugin-pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PineryIUSValue.java#L113) ). Looks like removal would have no impact.

mlaszloffy (approved):

PR #1224, opened 2023-07-31
Upgrade code formatter and apply new formatting

avarsava:
Spinoff from #1210 

POM and Maven JVM config as specified in https://github.com/Cosium/git-code-format-maven-plugin/blob/master/README.md


djcooke (approved):
It would be good to add a comment inside the plugin config in `pom.xml` about the additional config in `.mvn/jvm.config`

avarsava:
> It would be good to add a comment inside the plugin config in `pom.xml` about the additional config in `.mvn/jvm.config`

Done

callunity (approved):

PR #1227, opened 2023-08-11
Allow actions to know when they were last generated

apmasell:
Since some actions should change their behaviour if they are stale, this allows actions to know that. This information can't be used alone since `.actnow` actions will quickly become stale. This adds a separate policy to Vidarr that allows determining the desired submission behaviour to prevent stale actions from submitting.

JIRA Ticket: 

- [ ] Updates Changelog


plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/SubmitAction.java
```diff
@@ -149,11 +152,6 @@ public Stream<ActionCommand<?>> commands() {
     return state.canReattempt() ? Stream.of(DELETE, REATTEMPT, RESET) : Stream.of(RESET);
   }
 
-  @ActionParameter(name = "dry_run", required = false)
```
avarsava, line 152: Why is this removed? Pretty sure we use this option

avarsava (approved):

callunity (approved):

PR #1228, opened 2023-08-14
[GP-3706] docs

avarsava:
JIRA Ticket: GP-3706

Please let me know if anything is unclear. Some details are repeated because there wasn't a perfect place to put them. I'd love to put a big-picture walkthrough into some md dev docs here but i have yet to find a good way to do flow charts in md

- [N/A] Updates Changelog

### DIFF
--- a/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
+++ b/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
@@ -260,6 +260,20 @@ public final class Server implements ServerConfig, ActionServices {
             fileName ->
                 new GuidedMeditation(
                     fileName, DefinitionRepository.concat(definitionRepository, compiler)));
+
+    /**
+     * When fetch()ing a format, try the name of the format on:
+     * - AnnotatedInputFormatDefinition's Stream of known formats, which includes -input and -remote files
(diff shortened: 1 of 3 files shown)


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
```diff
@@ -260,6 +260,24 @@ public Server(int port, FileWatcher fileWatcher) throws IOException, ParseExcept
             fileName ->
                 new GuidedMeditation(
                     fileName, DefinitionRepository.concat(definitionRepository, compiler)));
+
+    /**
+     *
```
avarsava, line 265: How do I get the formatter to not put these two lines above a <pre> tag? if I remove them, it puts them back on commit.

callunity: Is there a particular reason that you want the two lines removed?

callunity (approved):
Very thorough - thanks for this clear explainer!

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/Server.java
callunity: You could see if there's a config setting to remove those two lines, but the formatter is pretty opinionated so that toggle might not exist.

grapigeau (approved):

PR #1229, opened 2023-08-16
Don't format javadoc

avarsava:
JIRA Ticket: N/A

Spinoff from #1228 because it keeps adding newlines to my javadoc

- [N/A] Updates Changelog

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -246,6 +246,11 @@
             <version>4.2</version>
           </dependency>
         </dependencies>
+        <configuration>
+          <formatterOptions>
+            <googleJavaFormat.formatJavadoc>false</googleJavaFormat.formatJavadoc>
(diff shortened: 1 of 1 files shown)


callunity (approved):

avarsava:
> Your other PR is avoiding standard JavaDoc formatting using `\<pre\>` tags for the entire body, which is not the intended use. The formatter plugin is probably adding blank lines because it thinks a JavaDoc should have an introductory sentence/paragraph before including HTML, which is valid.
> 
> I think having more standard JavaDocs is the better solution here, but it really doesn't matter, and this plugin is generally more headache than help anyway, so... 🤷‍♂️

Yeah, those `pre` tags were for getting it to preserve my formatting. i'll clean them up once this is merged

PR #1247, opened 2023-10-21
Replace JIRA client

apmasell:
Removes the unmaintained Atlassian-provided JIRA client with undesirable dependencies with a hand-written client.

JIRA Ticket: You don't want to know

- [X] Updates Changelog


mlaszloffy:
@apmasell can the merge conflict be resolved and then we'll start testing this. Thanks!

apmasell:
Rebased!

avarsava:
We're not going to be able to test the V3 type API I think. Should its untestedness be documented?

apmasell:
I've fully tested the V3 API on the cloud JIRA environment I got from Atlassian, so I think we're fully covered.

avarsava:
The plugin is able to reopen closed tickets, but the comment is not applied

apmasell:
It seems that the JIRA API doesn't support doing both in V2. I'll have to break them into two requests.

apmasell:
Okay, I split up the comment into two requests and it performs okay on V3. Let me know how V2 goes.

avarsava:
Should an action changing be able to update the assignee of an issue? I have an actnow that opens a ticket that assigns noone; if i purge the action and edit the actnow so it assigns me, should the regeneration of that action update the assignee?

apmasell:
The existing version didn't. I think it would be hard to know if you had reassigned the ticket and wouldn't want it to change the assignee if it reopened, right?

avarsava:
Could 'comment' in the OPEN block be called something like 'comment on reopen' or do the 'comment's in OPEN and CLOSE need to be named the same thing?

apmasell:
They don't need to be the same thing at all. Would `reopen_comment` work?

avarsava:
That'd be great, thanks

apmasell:
Done.

avarsava (approved):
I've tried and I can't seem to break anything else!

mlaszloffy (approved):
Looks great, thanks Andre

PR #1248, opened 2023-10-23
Add additional type guarantees

apmasell:
Create new type conversions for user-defined functions to allow consuming:

- dictionaries
- JSON data

JIRA Ticket: 

- [X] Updates Changelog

### DIFF
--- a/changes/add_tyguar.md
+++ b/changes/add_tyguar.md
@@ -1,0 +1,1 @@
+* Add new type conversion for dictionaries and JSON values in the Java API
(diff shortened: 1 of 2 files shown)


avarsava:
Do the language reference etc need to be updated?

apmasell:
No. These just allow custom functions to interact with existing Shesmu types. I had only written guarantees for things that were used and there were some missing gaps.

djcooke (approved):

PR #1249, opened 2023-10-23
Improve JavaDoc

apmasell:
Fix poorly formatted comments, add new ones, and add a GitHub workflow to build documentation

- [NA] Updates Changelog


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/plugins/PluginManager.java
```diff
@@ -1548,20 +1550,26 @@ protected void newInstance(GeneratorAdapter method) {
...
-     * PluginFileType) then the method is tracked as a static source. These input sources require no configuration.
-     * See implementation.md
+     * Perform validation on methods annotated with {@link ShesmuInputSource} and track as either a
+     * "static" or "dynamic" type of source. Validation on a {@link ShesmuInputSource} method is as
+     * follows:
      *
-     * @param method Method object of @ShesmuInputSource-annotated method to process
+     * <ol>
+     *   <li>Ensure method returns Stream
+     *   <li>Ensure method returns {@link Stream} or a subclass of the type parameter but not a
```
djcooke, line 1559: "Stream matching the type parameter?" Probably helpful to mention the location of the type parameter too (`FormatTypeWrapper`)?

djcooke (commented):
That's a lot of JavaDoc 😵‍💫

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/InputSource.java
```diff
@@ -1,31 +1,39 @@
...
+   *       data
+   *   <li>olives always request fresh data
+   *   <li>the simulator requests stale data by default, but can be set to request fresh data
+   * </ul>
    *
    * @param format name of the format requested
    * @param readStale false if fresh data requested from server
-   * @return Stream of Object where Object should be the input format data class
+   * @return Stream of the input format data class or a tuple containing all the fields in
+   *     alphabetical order
```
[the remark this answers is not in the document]
apmasell: Yes. I've clarified the comment.

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/plugins/PluginManager.java
apmasell: I've tried to clarify this. I'm not sure what you mean about `FormatTypeWrapper` as it's not used here.

apmasell:
Yes, well, there were a lot of "missing JavaDoc" complaints when I ran the tool.

I'll squash all this before merging.

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/plugins/PluginManager.java
djcooke: The method checks whether the method return type is `instanceof ParameterizedType`, and if not throws an exception with message

> Method %s of %s does not return Stream<T> or generic type information missing from bytecode.

and the docs originally referred to `Stream<T>` and `T`. Since the method is in the `FormatTypeWrapper` class, which has a type parameter `T`, I assumed this was the type parameter being referred to, but it seems that the method doesn't use any `T` at all, so unless I'm missing something, the exception message is wrong, but the updated JavaDoc is correct (and much clearer now, thanks!)

djcooke (approved):

djcooke:
Sorry, I didn't mean that as a complaint. It's a lot to look at, but great documentation! 🏆

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/plugins/PluginManager.java
apmasell: The exception isn't very helpful. The `T` is not referring to any `T` in this class. The issue here is how Java encodes generics... It's possible to write a method `Stream foo()` in a plugin which has no generic type information, and that's a problem (the one detected by line 1589). Maybe the exception should read `does not return Stream<> or generic type information missing`? Or maybe `Stream<...>`?

djcooke: Ah, makes sense. I think it is slightly clearer without the `T`

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/action/ActionState.java
```diff
@@ -62,10 +62,20 @@ public enum ActionState {
     this.processPriority = processPriority;
   }
 
+  /**
+   * The priority of an action state where the largest number indicates the most miserable action
```
avarsava, line 66: unclear - i thought smallest numbers were highest priority? i would want to act on the most miserable action first :(

apmasell: Now that I look, this is only used by the JIRA plugin and the replacement JIRA client won't use this. So, I say we delete this method after merging that. I'm happy to revert the comment change or we can just merge it and it goes away later. Your choice.

avarsava (dismissed):

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/filter/AlertFilter.java
```diff
@@ -300,16 +324,41 @@ private static <S> Parser parse3(
...
+   * Checks if the filter is inverted
+   *
+   * @return true if the filter should be inverted
+   */
   public boolean isNegate() {
     return negate;
   }
 
+  /**
+   * Convience method to construct a filter based on negation status
```
avarsava, line 346: Convenience

apmasell: Oh, let me know before you merge so I can squash this and any other changes.

avarsava (approved):

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/filter/AlertFilter.java
avarsava: squash pls!

PR #1251, opened 2023-11-06
Prepare compiler for modularization

apmasell:
The compiler currently emits classes that are incompatible with the Java 9 Jigsaw module system. This change refactors the compiler so that it generates module-compliant code in preparation for modularization. It does this by:

- making sure that all generated classes are in a single package
- adding a new method to `ActionGenerator` to provide a `MethodHandles.Lookup` instance

Under modularization, classes cannot reflect into classes outside of their own module visibility. For generated code, this makes it impossible for the compiler to rummage around in the class it just generated. To circumvent this, the generated class must create a `Lookup` instance with the generated class's visibility rules and then share this with the compiler.

- [NA] Updates Changelog (no visible changes)

### DIFF
--- a/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/runtime/ActionGenerator.java
+++ b/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/runtime/ActionGenerator.java
@@ -29,7 +29,9 @@ public abstract class ActionGenerator implements RequiredServices {
         }
 
         @Override
-        public Lookup privateLookup() {
+        public Lookup lookup() {
+          // Because this is a dummy instance, we're going to get the public lookup instance since
+          // there will be no exports that require using it.
           return MethodHandles.publicLookup();
         }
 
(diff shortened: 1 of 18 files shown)


shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/runtime/ActionGenerator.java
```diff
@@ -26,6 +28,11 @@ public Stream<String> inputs() {
           return Stream.empty();
         }
 
+        @Override
+        public Lookup privateLookup() {
+          return MethodHandles.publicLookup();
```
avarsava, line 33: 'Private lookup' returns a 'public lookup'?

apmasell: Yeah, poorly worded. I've renamed the method to `lookup` and added a clarification why this is public lookup in just this one instance.

avarsava (approved):

mlaszloffy (approved):

PR #1257, opened 2024-01-15
Redesign optional handling in `Group` clauses

apmasell:
Removes the `OnlyIf` collector and replace it with a more sophisticated set of collectors to handle optional types.

- [X] Updates Changelog


language.md
```diff
@@ -475,6 +469,37 @@ Performs multiple collections at once and converts the results into an object.
...
+
+- the unspecified behaviour is to simply drop empty optionals and proceed with
+  the collector as normal. This is equivalent to ``Where x != ` ` Min x Default
+  1234 ``; since `x` will never have the empty optional, the default will never
+  be used.
+- `OnlyIf All` requires that no empty input makes it to the collector; if any
+  optional input is found, the output from the collector is replaced by the
+  empty optional. So, if ``v = OnlyIf All `Min x?` `` was given `` `3` and `` ` `
+  ``, it would produce `` v == ` ` ``.
+- `OnlyIf Any` requires that one non-empty empty input makes it to the
```
avarsava, line 488: what is a 'non-empty empty input'? or is that a typo

apmasell: Typo.

language.md
```diff
@@ -475,6 +469,37 @@ Performs multiple collections at once and converts the results into an object.
...
+  `` ` ` ``, it would produce `` v == `3` ``, but `` ` ` `` and `` ` ` `` would
+  produce ``v == ` ` ``.
+- `Require All` requires that no empty input makes it to the collector; if any
+  optional input is found, the group will be rejected. So, if
+  ``v = Require All `Min x?` `` was given `` `3` and `` ` ` ``, it would not be
+  present in the output.
+- `Require Any` requires that one non-empty empty input makes it to the
+  collector; if no optional input is found, the group will be rejected. So, if
+  ``v = Require Any `Min x?` `` was given `` `3` and `` ` ` ``, it would produce
+  `v == 3`, but `` ` ` `` and `` ` ` `` would not be present in the output.
```
avarsava, line 500: why is `  ` twice here?

apmasell: I guess it doesn't have to be. I thought it was clearer to have matched numbers of inputs and outputs.

avarsava: but it is only once in the input?

apmasell: No, it could be multiple times in the input. Any number of empties would be treated a problem.

avarsava: I feel like i'm missing something. is this a rendering problem? there is one blank after 'was given' but two blanks in the example of what wouldn't be in the output 
image

apmasell: Oh, yeah, that's not how Jekyll renders it. Sigh. It's supposed to be like this ignoring linebreaks:

 ```
v = Require Any `Min x?`
```
 was given
```
 `3`
```
 and
```
 ` `
```
, it would produce
```
v == 3
```
, but
```
 ` `
```
 and
```
` `
```
would not be present in the output.

Comment (avarsava) [reply language.md]: Yeah, that's what the document says. Why is there 1 blank in the first clause and 2 blanks in the second?
I would think it would either be 'if v = Require Any Min x? was given 3, ' ', and ' ', it would produce 'v == 3' but ' ' and ' ' would not be present in the output'
or
'if v = Require Any Min x? was given 3 and ' ', it would produce 'v == 3' but ' ' would not be present in the output'

Comment (apmasell) [reply language.md]: I think I fixed the formatting. The second one is "only of blanks".

File: shesmu-server/src/test/resources/run/group-onlyif.shesmu [hunk 3]
```diff
@@ -1,6 +0,0 @@
-Version 1;
-Input test;
-
-Olive
- Group By workflow Into a = OnlyIf `project`
```
mlaszloffy, line 5: We use the pattern ``Group By workflow Into a = OnlyIf For i In `dictionary["key"]? As string`: Univalued i``

I'm looking at updating `OnlyIf` to `OnlyIf All` but it appears that `For` isn't supported with `OnlyIf All/Any` or at least not in the simulator?

apmasell: I'm assuming `dictionary[key]` is a single value, not a list, then the following should be equivalent:

```
Group By workflow Into a = `Univalued dictionary["key"]? As string`
```

avarsava (approved):

shesmu-server/src/test/resources/run/group-onlyif.shesmu
apmasell: Added a related test case that uses some JSON conversion.

mlaszloffy (approved):

PR #1260, opened 2024-01-15
Add `OnReject` support to `Group`

apmasell:
Allow `Group` clauses to take rejected groups and handle them in the same way as `Reject` and `Require`.

- [X] Updates Changelog

### DIFF
--- a/language.md
+++ b/language.md
@@ -215,8 +215,8 @@ dropped.
 
 This reshapes the data.
 
- - `Group` `By` _discriminator1_[`,` ...]  [`Where` _condition_] _collectionname1_ `=` _collector1_[`,` ...]
- - `Group` `By` _discriminator1_[`,` ...] `Using` _grouper_ _param_ `=` _expr1_[`,` ...]  [`With` _output_[`,` ...]]  [`Where` _condition_] _collectionname1_ `=` _collector1_[`,` ...]
+ - `Group` `By` _discriminator1_[`,` ...]  [`Where` _condition_] `Into` _collectionname1_ `=` _collector1_[`,` ...] [`OnReject` _reject1_[ _reject2_[ ...]] `Resume`]
+ - `Group` `By` _discriminator1_[`,` ...] `Using` _grouper_ _param_ `=` _expr1_[`,` ...]  [`With` _output_[`,` ...]]  [`Where` _condition_] `Into` _collectionname1_ `=` _collector1_[`,` ...] [`OnReject` _reject1_[ _reject2_[ ...]] `Resume`]
 
 Performs a grouping of the data. First, rows are collected in subgroups by
 their _discriminators_. If `Using` is provided, those subgroups are modified by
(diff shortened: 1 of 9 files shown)


language.md
```diff
@@ -245,6 +245,9 @@ another section. Each collector can have `Where` filters that limit the
 collected data. Optionally, a `Where` filter can be applied to all the
 collectors by providing _condition_.
 
+Rows which are rejected are passed to the rejection handlers. These are
+`Monitor` or `Dump` clauses or an `Alert` terminal. Rejection handler can only
```
avarsava, line 249: handler -> handlers

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/OliveNodeAlert.java
```diff
@@ -141,6 +142,7 @@ public void collectFreeVariables(Set<String> freeVariables) {
     freeVariables.add(SOURCE_LOCATION_LINE);
     freeVariables.add(SOURCE_LOCATION_COLUMN);
     freeVariables.add(SOURCE_LOCATION_HASH);
+    freeVariables.add("Olive Services");
```
avarsava, line 145: What does this do?

avarsava (approved):

shesmu-server/src/test/resources/run/group-onreject.shesmu
```diff
@@ -0,0 +1,11 @@
+Version 1;
+Input test;
+
+Olive
+ Group
+    By workflow
+    Into a = Univalued accession
+    OnReject
+      Alert alertname = "NotUnivalued", workflow = workflow, value = "true" For 15mins
```
mlaszloffy, line 9: I tried using accession in the alert but it looks like "Into" variables cannot be access in the OnReject scope? For this test/example, I was thinking it may be useful to list the values/accession that were expected to be univalued in the alert.

apmasell: That's correct. At least one of collected variables aren't in a valid state, so reading them would be undefined behaviour. Providing more information in this space is pretty constrained:

- the checks for each collector are organized in a short-circuiting chain; since the same chain is used for both handled and unhandled grouping problems, any extra work we do is going to happen for all grouping operations. Building two separate code paths would definitely add testing and complexity bugs. This makes collecting all the failed variable names not practical.
- the collected variables aren't organized in a sensible way, so if I tell you the first failure 1) it might not be the only failure and 2) it's not going to be consistent over slight changes to the olive, which violates the olive's goal of reproducibility
- the original records that were consumed by the grouper are gone, so I can't provide information about the input

More generally, what is the use case for Shesmu providing this information? Regardless of which collectors fail, a human needs to investigate the data anyway. It probably makes the most sense to provide a guided meditation to help investigate the failure.

Actually, thinking about that, maybe a better solution would be to create a reject handler that can trigger a guided meditation, so PL can have a meditation to-do list. Something to think about.

PR #1261, opened 2024-01-16
Allow JSON function arguments on _Definitions_ page

apmasell:
Provide a choice between Shesmu and raw JSON for function arguments in the _Try It_ dialog.

- [X] Updates Changelog

### DIFF
--- a/shesmu-server-ui/src/definitions.ts
+++ b/shesmu-server-ui/src/definitions.ts
@@ -1,4 +1,8 @@
-import { fetchJsonWithBusyDialog, locallyStored } from "./io.js";
+import {
+  fetchJsonWithBusyDialog,
+  locallyStored,
+  locallyStoredString,
+} from "./io.js";
 import {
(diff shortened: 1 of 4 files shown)


shesmu-server-ui/src/definitions.ts
```diff
@@ -689,12 +693,39 @@ function testFunction(func: FunctionDefinition): void {
...
   dialog((_close) => [
     "Test function ",
     func.name,
     br(),
+    "Argument Format: ",
+    dropdown(
+      (format, selected) => {
+        if (format == "json") {
+          return [
+            { type: "icon", icon: "braces" },
```
avarsava, line 708: Could these be icon + text? The icon-only dropdowns have proven confusing to users

apmasell: Done!

avarsava (approved):

rnaioicr (approved):

PR #1265, opened 2024-02-08
Add Vidarr retry failed provision-out command

apmasell:
Allows the new `/api/retry-provision-out` endpoint to be accessed via action command.

This PR will require an updated Vidarr dependency once https://github.com/oicr-gsi/vidarr/pull/342 is released.

- [X] Updates Changelog

### DIFF
--- a/plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/RunStateMonitor.java
+++ b/plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/RunStateMonitor.java
@@ -1,7 +1,10 @@
 package ca.on.oicr.gsi.shesmu.vidarr;
 
+import static ca.on.oicr.gsi.shesmu.vidarr.VidarrPlugin.MAPPER;
+
 import ca.on.oicr.gsi.shesmu.plugin.action.ActionState;
 import ca.on.oicr.gsi.vidarr.JsonBodyHandler;
+import ca.on.oicr.gsi.vidarr.api.RetryProvisionOutRequest;
 import ca.on.oicr.gsi.vidarr.api.SubmitWorkflowRequest;
 import ca.on.oicr.gsi.vidarr.api.WorkflowRunStatusResponse;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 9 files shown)


avarsava:
Hitting 'Retry Failed Workflow' is probably the most common thing a shesmu user does. Having to open the 'Commands' dropdown adds an extra click. Would it be possible to determine which retry button to display based on the operations information?

apmasell:
Sure. We could give commands a priority and then display the top 3 and bury the rest. Does that sound reasonable? Out of scope for this PR though.

avarsava (approved):

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/RunStateMonitor.java
```diff
@@ -88,20 +93,35 @@ public RunStateMonitor(String workflowRunUrl, WorkflowRunStatusResponse status)
...
+    // Failed during provision out may be retried
+    if (status.getEnginePhase() != null
+        && (status.getEnginePhase().equals("FAILED")
+            && status.getOperations() != null
+            && status.getOperations().stream()
+                .anyMatch(
+                    operation ->
+                        operation.getEnginePhase().equals("PROVISION_OUT")
+                            && operation.getStatus().equals("FAILED")))) {
+      return AvailableCommands.CAN_RETRY;
```
mlaszloffy, line 106: While testing this PR, it looks like when a workflow run fails at provision out, both the "Reattempt Failed Workflow" and "Retry Provision Out" buttons are being displayed. Is this expected? It looks like here the buttons that should be displayed are being restricted to the relevant commands.

apmasell: Sure. I mean, is there a state when it can retry and there's a reason to reattempt? In #1272 I'm adding a ranking to the commands, so we can have retry is displayed first. If there's no good reason to reattempt, I'll change it.

mlaszloffy: I can't think of a reason to reattempt (start the Vidarr workflow run from the beginning) when a Vidarr workflow run can be retried (failed at provision out).

Understood that the other PR adds ranking of retry and reattempt commands/operation/buttons, but it looks like here the available commands should be restricted to "CAN_RETRY", but I'm seeing action failed at provision file out with both retry and reattempt as available commands.

Another consideration is that failed actions can be worked on in bulk. It does not appear that the bulk "Reattempt Failed Workflow" and "Retry Provision Out" operation/buttons categorize the selected actions into can/should reattempt vs. can/should retry.

apmasell: Okay. Fixed.

mlaszloffy: Thanks Andre, looks great!

mlaszloffy (approved):

PR #1274, opened 2024-03-05
New fix for 'Cannot cast Optional to Set'

avarsava:
Result of `git bisect`. Tested that casting exception does not occur.

- [ X] Updates Changelog


mlaszloffy (approved):

avarsava:
We have an actual fix for this now, no need to revert features.

PR #1290, opened 2024-04-17
[GP-4245] Debug logging for jira client

avarsava:
JIRA Ticket: GP-4245

- [X] Updates Changelog


avarsava:
> Probably want to combine multi-part messages into a single message to avoid them getting separated since multiple threads will be writing to the same log

Makes sense! fixed

djcooke (approved):

jrajawat (approved):

avarsava:
Re-requesting review because I realized nothing had a proper toString so the debug logs would be less than useful and then everything spiraled out of control from there

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/cache/KeyValueCache.java
```diff
@@ -169,4 +169,30 @@ public final void ttl(int ttl) {
...
+        .append(", records = ")
+        .append(records.entrySet())
+        .append(", ttl = ")
+        .append(ttl);
+    return writeOut.toString();
+  }
+
+  protected abstract String getKType();
+
+  protected abstract String getVType();
```
djcooke, line 197: `getKeyType` and `getValueType` would be more clear, but you're also basically defining these at the class level. Would it be more useful to just log the class name in `toString` rather than the type parameters? That would also be more concise and impossible to mess up when implementing a new subclass

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/filter/ActionFilterType.java
```diff
@@ -30,4 +30,14 @@ public String[] getTypes() {
   public void setTypes(String[] types) {
     this.types = types;
   }
+
+  @Override
+  public String toString() {
+    StringBuilder writeOut = new StringBuilder();
+    writeOut.append("Action type filter of types: ");
+    for (String type : types) {
+      writeOut.append(type).append(", ");
```
djcooke, line 39: Leaves an extra comma at the end, which could be confusing. Could use String.join instead, or delete from the StringBuilder after. Same pattern in a few other places too

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/filter/BaseAgoActionFilter.java
```diff
@@ -42,4 +42,13 @@ public final long getOffset() {
...
   }
+
+  @Override
+  public String toString() {
+    StringBuilder writeOut = new StringBuilder();
+    writeOut.append(getOperation()).append(" filter of offset: ").append(offset);
+    return writeOut.toString();
+  }
+
+  protected abstract String getOperation();
```
djcooke, line 53: Would logging the class name here be just as useful too since these are all defined basically at the class level?

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/filter/BaseRangeActionFilter.java
```diff
@@ -69,4 +69,19 @@ public final void setEnd(Long end) {
   public final void setStart(Long start) {
     this.start = start;
   }
+
+  protected abstract String getName();
```
djcooke, line 73: Similarly, would the logging the class name be just as good?

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/cache/KeyValueCache.java
avarsava: I designed it this way to minimize repeat code, if I moved toString to the class level then there would be ~20 lines of repeat code in each child class of KeyValueCache

djcooke: I meant that "KType" and "VType" are being defined at the class level right now - the return value is a static String for all implementations. This means the class name is just as useful, if not more so since 2 classes could have the same types. In `KeyValueCache#toString` you could just log `getClass()` (or one of its String properties) instead of needing `getKType` and `getVType` methods

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/filter/ActionFilterType.java
avarsava: Fixed

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/cache/KeyValueCache.java
avarsava: Didn't realize this would get the actual class name instead of the abstract class's name, thanks, fixed

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/filter/BaseAgoActionFilter.java
avarsava: Fixed

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/filter/BaseRangeActionFilter.java
avarsava: Fixed

djcooke:
Looks good once compile errors are fixed (missed removing a few of the getKType/getVType/getOperation implementations)

jrajawat (approved):

PR #1291, opened 2024-04-19
Fix authentication using ShAWK client

apmasell:
Corrects a bug where authentication would discard the password from the user and send a passwordless `Basic` header and allows the user to try again if authorization fails.

JIRA Ticket: 

- [X] Updates Changelog


avarsava (approved):
Tested and it works

mlaszloffy (approved):

PR #1314, opened 2024-07-11
[GP-4316] Use double quotes to match exact summary

avarsava:
JIRA Ticket: GP-4316

Thanks @mlaszloffy for finding this workaround

- [X] Updates Changelog


avarsava:
Actually it's better than nothing, let's get this in and then also I'll do the tagging thing

mlaszloffy (approved):

PR #1331, opened 2024-10-16
Gp 4415.add cardea detailed case input format

grapigeau:
JIRA Ticket: https://jira.oicr.on.ca/browse/GP-4415

- [x] Updates Changelog

### DIFF
--- a/changes/add_cardea_detailed_case_input_format.md
+++ b/changes/add_cardea_detailed_case_input_format.md
@@ -1,0 +1,1 @@
+Cardea input format for detailed case summaries


avarsava:
Detailed summaries seem to differ from the original summaries by three columns - why not expand the existing format?

grapigeau:
> Detailed summaries seem to differ from the original summaries by three columns - why not expand the existing format?
The eventual goal is to deprecate and replace the existing case_summary but the understanding is this will not be an instant transition.
The meeting notes linked in the ticket give a little more info: https://wiki.oicr.on.ca/display/GSI/Review+of+New+Case+Format+for+Shesmu

avarsava:
If none of the columns in use in the case_summary format are changing, then adding new columns to the format won't affect production olives at all. I'm not sure I understand the concern

grapigeau:
> If none of the columns in use in the case_summary format are changing, then adding new columns to the format won't affect production olives at all. I'm not sure I understand the concern

We decided not to include limsIds (list) in Cardea since it's duplicate information so it's not just an addition of columns.

avarsava:
What do you mean by limsIds not being in cardea? Where do they come from then?

grapigeau:
> What do you mean by limsIds not being in cardea? Where do they come from then?

I mean the value `"limsIds": [
      "string"
    `],`` which is present at the /shesmu-cases endpoint is not available at the /shesmu-detailed-cases endpoint. I should've specified new Cardea endpoint, not "Cardea."

mlaszloffy:
> Detailed summaries seem to differ from the original summaries by three columns - why not expand the existing format?

caseStatus logic is also changing.

mlaszloffy (approved):

avarsava (approved):

PR #1340, opened 2025-01-08
add support for ignore versions in cerberus plugin

grapigeau:
Result of https://github.com/oicr-gsi/cerberus/commit/bf695bbb606fb2a0e57c64b8eeefe4d30ff82f25#diff-d87c4afe90c4ee378625bb08cfb2a527e58c388b50b8adfef13d8da0a8381cefR81

JIRA Ticket: [GP-4493](https://jira.oicr.on.ca/browse/GP-4493)

- [ ] Updates Changelog

### DIFF
--- a/plugin-cerberus/src/main/java/ca/on/oicr/gsi/shesmu/cerberus/CerberusPlugin.java
+++ b/plugin-cerberus/src/main/java/ca/on/oicr/gsi/shesmu/cerberus/CerberusPlugin.java
@@ -183,6 +183,7 @@ public final class CerberusPlugin extends JsonPluginFile<Configuration> {
                                 entry.getKey(),
                                 entry.getValue().getUrl(),
                                 entry.getValue().getVersions()))));
+    final List<String> ignoreVersions = configuration.getIgnore();
     final var versions =
         configuration.getPinery().values().stream()
             .flatMap(pc -> pc.getVersions().stream())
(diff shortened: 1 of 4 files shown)


plugin-cerberus/src/main/java/ca/on/oicr/gsi/shesmu/cerberus/CerberusPlugin.java
```diff
@@ -183,6 +183,7 @@ protected Optional<Integer> update(Configuration configuration) {
                                 entry.getKey(),
                                 entry.getValue().getUrl(),
                                 entry.getValue().getVersions()))));
+    final List<String> ignoreVersions = configuration.getIgnore();
```
avarsava, line 186: Does this need a change to https://github.com/oicr-gsi/shesmu/blob/master/docs/plugin-cerberus.md ?

avarsava:
In addition, the cerberus code should probably protect against null ignoreProviders, probably with a Nonnull annotation in the constructor and then maybe also detecting null and forcing an empty list instead (or refusing to start the plugin with invalid config if these are super important)

avarsava (approved):

grapigeau:
> In addition, the cerberus code should probably protect against null ignoreProviders, probably with a Nonnull annotation in the constructor and then maybe also detecting null and forcing an empty list instead (or refusing to start the plugin with invalid config if these are super important)

I think forcing an empty list instead makes the most sense here and matches the intent in [Cerberus](https://github.com/oicr-gsi/cerberus/blob/master/cerberus-cli/src/main/java/ca/on/oicr/gsi/cerberus/cli/RunOnline.java#L52)

callunity (approved):

wuall826 (approved):

PR #1366, opened 2025-05-16
[GP-3110] remotejson timeouts

avarsava:
JIRA Ticket: GP-3110

also devarifies the file. easier to read if you only look at the timeout commit

- [X] Updates Changelog

### DIFF
--- a/docs/input-formats.md
+++ b/docs/input-formats.md
@@ -25,11 +25,14 @@ To access data remotely, create a file ending in `.`_format_`-remote` as follows
     {
        "authentication": null,
        "url": "http://some.url/format/endpoint",
-       "ttl": 10
+       "ttl": 10,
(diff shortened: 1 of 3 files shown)


mlaszloffy (approved):

docs/input-formats.md
```diff
@@ -25,11 +25,14 @@ To access data remotely, create a file ending in `.`_format_`-remote` as follows
     {
        "authentication": null,
        "url": "http://some.url/format/endpoint",
-       "ttl": 10
+       "ttl": 10,
+       "timeout": 60
```
djcooke, line 29: Just looking at the json, I'd assume this is seconds, especially with the example of 60 (is setting this to an hour reasonable?). Might want to clarify the field name and/or the example.

PR #1370, opened 2025-06-27
add in extra check for 'null' in case_archive metadata to account for json parsing behaviour

grapigeau:
JIRA Ticket: https://jira.oicr.on.ca/browse/GP-4751

- [ ] Updates Changelog


callunity (approved):

plugin-nabu/src/main/java/ca/on/oicr/gsi/shesmu/nabu/NabuPlugin.java
```diff
@@ -193,10 +193,10 @@ public String NabuToken() throws RuntimeException {
   private Optional<String> findMetadataTextValue(
       JsonNode metadata, String bestKey, String backupKey) {
     // metadata for earlier case archives were submitted with snake case keys
-    if (metadata.findValue(bestKey) != null) {
+    if (metadata.findValue(bestKey) != null && !metadata.findValue(bestKey).equals("null")) {
       return Optional.of(metadata.findValue(bestKey).textValue());
     }
-    if (metadata.findValue(backupKey) != null) {
+    if (metadata.findValue(backupKey) != null && !metadata.findValue(bestKey).equals("null")) {
```
callunity, line 199:
```suggestion
    if (metadata.findValue(backupKey) != null && !metadata.findValue(backupKey).equals("null")) {
```

callunity (approved):

plugin-nabu/src/main/java/ca/on/oicr/gsi/shesmu/nabu/NabuPlugin.java
```diff
@@ -193,11 +193,12 @@ public String NabuToken() throws RuntimeException {
   private Optional<String> findMetadataTextValue(
       JsonNode metadata, String bestKey, String backupKey) {
     // metadata for earlier case archives were submitted with snake case keys
-    if (metadata.findValue(bestKey) != null) {
-      return Optional.of(metadata.findValue(bestKey).textValue());
+    if (metadata.findValue(bestKey) != null && metadata.findValue(bestKey).textValue() != null) {
```
avarsava, line 196:
```suggestion
    if (metadata.findValue(bestKey) != null) {
```
If you're using Optional.ofNullable, it will create an Optional.empty if metadata.findValue(bestKey).textValue() is null. 
Ditto for line 199.

callunity:
> Upstream should be fixed to not serialize "null" when it means `null`. Otherwise you have to implement this behaviour in everything that ingests that endpoint, which might not only be shesmu in the future. And evidently it can be missed the first time!

We looked into this further and found that "null" was a one-off anomaly and not a systemic serialization issue in Nabu

PR #1379, opened 2025-10-02
Patch for null filename, but why is filename null?

avarsava:
JIRA Ticket: n/a

Patch for #1376 

As for why this happens: backlogged GP-4906

- [n/a] Updates Changelog
- [n/a] Updates developer documentation


callunity (approved):
From the [class Javadoc](https://github.com/oicr-gsi/shesmu/blob/83ebec3f4befc4847ddb9b1aed64413372b60e76/shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/compiler/definitions/ConstantDefinition.java#L37C87-L37C90), it sounds like some constants come from Java libraries and so aren't backed by a file:
> `They aren't constant in the sense that they can be arbitrary bytecode, so <code>now</code> is  considered a constant even though it varies`

PR #1418, opened 2026-03-24
Fix code generation with aliased signatures

apmasell:
Addresses a bug where signatures that were aliased by an import would generate incorrect code. Shesmu only handles things by name, so when `Import std::*;`, there will be two signatures `std::signature::json` and `signature::json`. Some code paths would generate both versions and some only one. This bug was discovered with a `Dump All` since it iterates over all possible stream variables, but a user could just have easily written `signature::json`. This adds checks in multiple places to only process unaliased signatures.

JIRA Ticket: 

- [X] Updates Changelog
- [X] Updates developer documentation (or N/A)

### DIFF
--- a/changes/fix_import_signature.md
+++ b/changes/fix_import_signature.md
@@ -1,0 +1,1 @@
+* Fixes multiple bugs where `Import std::*` would create duplicate, broken signatures for `std::signature::json` and friends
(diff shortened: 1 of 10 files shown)


avarsava (approved):
Thank you!

callunity (approved):

PR #1419, opened 2026-04-02
[GP-5046] Import actions

avarsava:
JIRA Ticket: GP-5046

Sister PR: https://github.com/oicr-gsi/vidarr/pull/464
Will unset the pom.xml changes when that PR is released

Open to changing the `import_` thing to being named differently. It'd be kind of messy if Analysis really named something starting with `import_`. In the future maybe allowing imports could be toggled at the workflow level or something

- [X] Updates Changelog


avarsava:
Ah, it helps if you actually mark TODOs with TODO so the IDE tells you about it. I'll look at my refactor notes next week

avarsava:
> Ah, it helps if you actually mark TODOs with TODO so the IDE tells you about it. I'll look at my refactor notes next week

Addressed these with 1bbbf74baf92a0aea6d7f21c5b2f34dd6be6dfe2

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/ImportAction.java
```diff
@@ -0,0 +1,230 @@
...
+
+public class ImportAction extends VidarrAction {
+  ImportRequest request;
+  ImportState state = new ImportStateAttemptSubmit();
+
+  public ImportAction(Supplier<VidarrPlugin> owner, WorkflowDeclaration workflow) {
+    super("vidarr-import", owner);
+
+    this.request = ImportRequest.fromDeclaration(workflow);
+    request.getWorkflowVersion().setWorkflow("");
```
callunity, line 36: Why set to empty string?

avarsava:
Test failures are some docker nonsense?

callunity:
> Test failures are some docker nonsense?

I've updated the repo settings so the Docker actions can now run. Now the failures are just due to the -SNAPSHOT version. Vidarr release soon!

callunity:
> Open to changing the import_ thing to being named differently. It'd be kind of messy if Analysis really named something starting with import_. In the future maybe allowing imports could be toggled at the workflow level or something

Fastqs already get injected as `import_fastq` workflows... Perhaps `load_` ? (There's the concern about semantic overload with Vidarr's `/load` endpoint, but with importing being load+reprovision perhaps it is semantically relevant enough to work?)

avarsava:
> (There's the concern about semantic overload with Vidarr's /load endpoint, but with importing being load+reprovision perhaps it is semantically relevant enough to work?)

I share the concern about semantic overload. Load, unload, and reprovision seem poorly understood already, using `load_` to mean import here would probably worsen the confusion.

Would vidarr /import-ing make `import_fastq` redundant? Could this be a short-lived problem?

Alternatively, is the `_` the problem? Would making a pseudo namespace like `import::` instead be better?

callunity:
> Would vidarr /import-ing make `import_fastq` redundant? Could this be a short-lived problem?
 
Unfortunately, I don't know that it will - it's `import_fastq` because if the files are coming from collaborators, we don't know what workflow produced the files (could be bcl2fastq, or bclconvert) and I think we still need to plan for loading in files from collaborators.

> Alternatively, is the `_` the problem? Would making a pseudo namespace like `import::` instead be better?

I think this is a great idea!

avarsava:
> I think this is a great idea!

Done! The shesmu UI even takes it at face value and puts them in their own little section in the Definitions guide.

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/VidarrPlugin.java
```diff
@@ -545,6 +553,91 @@ public void store(
                                                                     workflow.getName()))));
                                               }
                                             })));
+
+                // Create import actions
```
callunity, line 557: I think there _might_ be a future wherein we'd want to allow imports on a Vidarr that doesn't accept submissions (ie. importing to archival, though I haven't worked through all the ramifications yet). What do you think about enabling this by changing this: 
```
if !canSubmit 
    (exit early)
else
    (add submit and import actions)
```
 to 
```
if canSubmit
    (add submit actions)
add import actions
```
?

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/ImportStateMonitor.java
```diff
@@ -0,0 +1,130 @@
...
+            HttpRequest.newBuilder(url).GET().build(),
+            new JsonBodyHandler<>(MAPPER, WorkflowRunStatusResponse.class));
+    if (response.statusCode() == 200) {
+      final WorkflowRunStatusResponse result = response.body().get();
+
+      // Translating operation statuses to action states is the same for both modes
+      final ActionState status = RunStateMonitor.actionStatusForWorkflowRun(result);
+      if (status == ActionState.FAILED) {
+        return new PerformResult(
+            List.of("Reprovisioning has failed while executing. See provisioner logs for details."),
```
callunity, line 47: `Reprovisioning` -> `Importing` since the pipeline leads will be thinking of it as an `import` action?
`provisioner logs` -> `vidarr logs` ?

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/VidarrPlugin.java
avarsava: I see no issues, but it's also an easy change to do later if you want to think through the ramifications first

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/ImportAction.java
avarsava: Great question. I don't remember why I did that

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/ImportStateMonitor.java
avarsava: If we have gotten far enough into the importing process that we're able to be monitoring, it's reprovisioning that has failed, not loading. The output provisioner logs (eg the cromwell logs, in the case of using a CromwellOutputProvisioner) will have much more information than vidarr's logs could.

avarsava: (a failure to load would have failed in ImportStateAttemptSubmit)

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/ImportAction.java
avarsava: Ah, right. We don't have the full text of the workflow when creating an import from information available in an olive, and we agreed we didn't need it. The string representation of the workflow is set to an empty string for the injection mode of loading on vidarr's side. Otherwise there's a null pointer exception

callunity: That sounds reasonable. Would you mind adding some of that as a comment here?

avarsava: Done!

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/VidarrPlugin.java
avarsava: Did you want to wait on this or do it now? As it is, it's less permissive, so it's safe to leave as is

callunity: ok to wait

callunity (approved):

grapigeau (approved):

PR #1434, opened 2026-06-01
Don't require array of version hashes in import external keys

avarsava:
JIRA Ticket: N/A

Make it match the submit action

- [X] Updates Changelog

### DIFF
--- a/plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/ImportAction.java
+++ b/plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/ImportAction.java
@@ -117,7 +117,7 @@ public class ImportAction extends VidarrAction {
   }
 
   @SuppressWarnings("unchecked")
-  @ActionParameter(name = "external_keys", type = "ao4id$sprovider$sstale$bversions$msas")
+  @ActionParameter(name = "external_keys", type = "ao4id$sprovider$sstale$bversions$mss")
   public void externalKeys(Set<Tuple> values) {
     request
         .getWorkflowRun()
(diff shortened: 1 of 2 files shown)


plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/ImportAction.java
```diff
@@ -129,7 +129,14 @@ public void externalKeys(Set<Tuple> values) {
                       externalKey.setId((String) value.get(0));
                       externalKey.setProvider((String) value.get(1));
                       stale |= (Boolean) value.get(2);
-                      externalKey.setVersions((Map<String, Set<String>>) value.get(3));
+
+                      // IntelliJ thinks this should be ConcurrentHashMap?
```
callunity: I don't think this operates in a threaded context so probably not necessary?

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/ImportAction.java
```diff
@@ -129,7 +129,14 @@ public void externalKeys(Set<Tuple> values) {
...
+                      Map<String, Set<String>> versions = new HashMap<>();
```
callunity, line 133: Ah, I see - I was trying to resolve this issue with this Vidarr PR but stopped work on it before completing: https://github.com/oicr-gsi/vidarr/pull/430
That's why the SubmitAction works behind the scenes to convert the external keys from the olive to `Collection<ExternalKey>` but the ImportAction works behind the scenes to convert the external keys from the olive to `Collection<ExternalMultiVersionKey>` (because its output needs to look like `ProvenanceWorkflowRun<ExternalMultiVersionKey>`). 
I'll reference this PR in the Vidarr PR and see about revisiting it.

callunity (approved):

grapigeau (approved):

PR #1451, opened 2026-08-28
Upgrade to Jackson 3

callunity:
JIRA Ticket: GP-5743

In draft now as it needs updated releases of Cerberus and Vidarr.

Includes tests for the changes to ExtractorScriptNode, where the bytecode generation changed.

- [x] Updates Changelog


plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/VidarrAction.java
```diff
@@ -25,12 +25,12 @@ public abstract class VidarrAction extends Action {
           .asList();
   protected boolean stale;
 
-  protected static boolean checkJson(JsonNode json, Pattern query) {
+  protected static boolean checkJson(ObjectNode json, Pattern query) {
```
avarsava: Why this change? Doesn't this rule out the whole `case ARRAY:` block?

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/ImportAction.java
```diff
@@ -228,7 +228,7 @@ public boolean search(Pattern query) {
                                             .anyMatch(value -> query.matcher(value).matches())))
         || checkJson(request.getWorkflowRun().getArguments(), query)
         || checkJson(request.getWorkflowRun().getMetadata(), query)
-        || checkJson(request.getWorkflowRun().getEngineParameters(), query)
+        || checkJson((ObjectNode) request.getWorkflowRun().getEngineParameters(), query)
```
avarsava: Should engine parameters change to being explicitly an ObjectNode on Vidarr's side, rather than it being ambiguous but we force it here?

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/ImportAction.java
```diff
@@ -251,14 +251,16 @@ public Stream<String> tags() {
...
+  public ObjectNode toJson(JsonMapper mapper) {
     ObjectNode node = super.toJson(mapper);
 
     // Need to convertValue with the Mapper in order to use the Java Time parsing
     node.set("request", mapper.convertValue(request, JsonNode.class));
     // Bring these up for the 'Differences from Olive' sections
-    ((ObjectNode) node.get("request")).set("arguments", request.getWorkflowRun().getArguments());
-    ((ObjectNode) node.get("request")).set("metadata", request.getWorkflowRun().getMetadata());
+    ((ObjectNode) node.get("request"))
+        .set("arguments", (JsonNode) request.getWorkflowRun().getArguments());
```
avarsava: getArguments returns ObjectNode, which is extends JsonNode. Why the cast here?

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/ImportAction.java
```diff
@@ -251,14 +251,16 @@ public Stream<String> tags() {
...
-  public ObjectNode toJson(ObjectMapper mapper) {
+  public ObjectNode toJson(JsonMapper mapper) {
     ObjectNode node = super.toJson(mapper);
 
     // Need to convertValue with the Mapper in order to use the Java Time parsing
     node.set("request", mapper.convertValue(request, JsonNode.class));
     // Bring these up for the 'Differences from Olive' sections
-    ((ObjectNode) node.get("request")).set("arguments", request.getWorkflowRun().getArguments());
-    ((ObjectNode) node.get("request")).set("metadata", request.getWorkflowRun().getMetadata());
+    ((ObjectNode) node.get("request"))
```
avarsava: Should these casts change to 
```suggestion
    node.get("request").asObject()
```
now that that's available?

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/core/actions/fake/LocalFile.java
```diff
@@ -25,19 +25,19 @@ public void configuration(SectionRenderer renderer) {}
...
-      var name = obj.get("name").asText();
+      var name = obj.get("name").asString();
       definer.defineAction(
           name,
-          "Fake version of: " + obj.get("description").asText(),
+          "Fake version of: " + obj.get("description").asString(),
           FakeAction.class,
           () -> new FakeAction(name),
-          Utils.stream(obj.get("parameters").elements())
+          Utils.stream(obj.path("parameters").values())
```
avarsava: why path in this case? wouldn't the get()s 4 lines down still cause exceptions for the missing node?

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/ActionProcessor.java
```diff
@@ -373,7 +373,8 @@ public String name() {
           return "external";
         }
       };
-  private static final JsonNodeFactory JSON_FACTORY = JsonNodeFactory.withExactBigDecimals(false);
```
avarsava, line 376: Has this become a default setting?

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/ImportAction.java
callunity: yep, is unnecessary cast

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/ImportAction.java
callunity: Nice, I had been going off the Jackson 2.x->3.0 migration docs and that function wasn't available until 3.1. Will update!

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/VidarrAction.java
callunity: I'm not actually sure. Will revert.

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/server/ActionProcessor.java
callunity: This setting was never actually getting used. The setting got moved away from the JsonNodeFactory and onto the mapper's builder, and since the mapper for this class comes from RuntimeSupport it was going to be complicated to rebuild, so I went to see if the factory had actually been used for BigDecimals. The only use of `JSON_FACTORY.numberNode()` was line 397, where it's operating on a `long`, not on a BigDecimal, so there was no need to add the functionality.

shesmu-server/src/main/java/ca/on/oicr/gsi/shesmu/core/actions/fake/LocalFile.java
callunity: I'm not sure what happened here; will fix.

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/ImportAction.java
avarsava: Although, I guess this is irrelevant if the signature for checkJson goes back to using JsonNode. Nevermind!

avarsava (approved):
Thank you!

plugin-vidarr/src/main/java/ca/on/oicr/gsi/shesmu/vidarr/ImportAction.java
callunity: engineParameters was set to JsonNode to be compatible with the WorkflowEngine API, as explained in this commit: https://github.com/oicr-gsi/vidarr/commit/35643564
the WorkflowEngine 's `engineParameters` is of type `Optional<BasicType>` so it's not restricted to objects.

grapigeau (approved):

PR #1453, opened 2026-09-01
add runscanner::miso::ultima_workflowfiles function to get the analys…

grapigeau:
…is output files for an Ultima run

JIRA Ticket: 

- [x] Updates Changelog

### DIFF
--- a/plugin-runscanner/src/main/java/ca/on/oicr/gsi/shesmu/runscanner/RunScannerClient.java
+++ b/plugin-runscanner/src/main/java/ca/on/oicr/gsi/shesmu/runscanner/RunScannerClient.java
@@ -6,6 +6,9 @@ import ca.on.oicr.gsi.runscanner.dto.dragen.DragenAnalysisUnit;
 import ca.on.oicr.gsi.runscanner.dto.dragen.DragenPipelineRun;
 import ca.on.oicr.gsi.runscanner.dto.dragen.DragenWorkflowRun;
 import ca.on.oicr.gsi.runscanner.dto.type.IlluminaChemistry;
+import ca.on.oicr.gsi.runscanner.dto.ultima.UltimaAnalysisUnit;
+import ca.on.oicr.gsi.runscanner.dto.ultima.UltimaPipelineRun;
+import ca.on.oicr.gsi.runscanner.dto.ultima.UltimaWorkflowRun;
 import ca.on.oicr.gsi.shesmu.plugin.cache.InitialCachePopulationException;
 import ca.on.oicr.gsi.shesmu.plugin.functions.ShesmuMethod;
 import ca.on.oicr.gsi.shesmu.plugin.functions.ShesmuParameter;
@@ -79,14 +82,9 @@ public final class RunScannerClient extends JsonPluginFile<Configuration> {
               : Optional.empty();
 
       if (run instanceof IlluminaNotificationDto) {
-        List<PipelineRun> sortedPipelineRuns = ((IlluminaNotificationDto) run).getPipelineRuns();
-        if (sortedPipelineRuns.isEmpty()) {
-          pipelineRun = Optional.empty();
-        } else {
-          // Plugin only serves the latest analysis
-          sortedPipelineRuns.sort(Comparator.comparingInt(PipelineRun::getAttempt));
-          pipelineRun = Optional.of(sortedPipelineRuns.get(sortedPipelineRuns.size() - 1));
-        }
+        pipelineRun = latestPipelineRun(((IlluminaNotificationDto) run).getPipelineRuns());
+      } else if (run instanceof UltimaNotificationDto) {
+        pipelineRun = latestPipelineRun(((UltimaNotificationDto) run).getPipelineRuns());
       } else {
         pipelineRun = Optional.empty();
       }
@@ -159,6 +157,19 @@ public final class RunScannerClient extends JsonPluginFile<Configuration> {
     return RunScannerPluginType.getFlowcellLayout(run.getLaneCount(), isJoined);
   }
 
+  /**
+   * Plugin only serves the latest analysis attempt
+   *
+   * @param pipelineRuns the pipeline run attempts for a run
+   */
+  private static Optional<PipelineRun> latestPipelineRun(List<PipelineRun> pipelineRuns) {
+    if (pipelineRuns.isEmpty()) {
+      return Optional.empty();
+    }
+    pipelineRuns.sort(Comparator.comparingInt(PipelineRun::getAttempt));
+    return Optional.of(pipelineRuns.get(pipelineRuns.size() - 1));
+  }
+
   private Instant lastUpdate = Instant.EPOCH;
   private final ProgressiveRequestDto request = new ProgressiveRequestDto();
   private final Map<String, RunInformation> runCache = new ConcurrentHashMap<>();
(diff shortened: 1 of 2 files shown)


avarsava:
Isn't there more than one workflow on Ultima? What's the envisioned use for getting all of them at once?

grapigeau:
> Isn't there more than one workflow on Ultima? What's the envisioned use for getting all of them at once?

There is. I figured the workflow would be passed in as a parameter (rather than having different import functions) since the functionality is the same whether it's a cram generation or emseq metadata workflow, just the files themselves will be different

plugin-runscanner/src/main/java/ca/on/oicr/gsi/shesmu/runscanner/RunScannerClient.java
```diff
@@ -231,6 +242,34 @@ public Optional<JsonNode> dragen_bclconvert(
     }
   }
 
+  @ShesmuMethod(description = "Get the analysis files for an Ultima run")
+  public Optional<JsonNode> ultima_workflowfiles(
+      @ShesmuParameter(description = "name of run") String run,
+      @ShesmuParameter(description = "barcode, formatted like ATCGATCGATCGATCG") String barcode,
+      @ShesmuParameter(description = "workflow name") String workflow,
+      @ShesmuParameter(description = "library") String library) {
+    try {
```
avarsava, line 252: Sorry if I missed something along the way, but shouldn't a workflow file be identifiable by IUS since libraries can be repeated across lanes?

plugin-runscanner/src/main/java/ca/on/oicr/gsi/shesmu/runscanner/RunScannerClient.java
```diff
@@ -231,6 +242,34 @@ public Optional<JsonNode> dragen_bclconvert(
     }
   }
 
+  @ShesmuMethod(description = "Get the analysis files for an Ultima run")
+  public Optional<JsonNode> ultima_workflowfiles(
+      @ShesmuParameter(description = "name of run") String run,
+      @ShesmuParameter(description = "barcode, formatted like ATCGATCGATCGATCG") String barcode,
```
avarsava, line 248: No `-`?

plugin-runscanner/src/main/java/ca/on/oicr/gsi/shesmu/runscanner/RunScannerClient.java
```diff
@@ -231,6 +242,34 @@ public Optional<JsonNode> dragen_bclconvert(
...
+      @ShesmuParameter(description = "workflow name") String workflow,
```
avarsava: If you're going with a 'one function for all workflows' approach, I worry about the discoverability of what valid workflow names here are. If there will eventually be a lot, it'd be a pain to update this description with them all, but where else could they go?

plugin-runscanner/src/main/java/ca/on/oicr/gsi/shesmu/runscanner/RunScannerClient.java
avarsava: Also, if run+barcode+library is the unique identifier, then imo workflow should be either first or last in the parameters

plugin-runscanner/src/main/java/ca/on/oicr/gsi/shesmu/runscanner/RunScannerClient.java
grapigeau: Yeah that's a very fair point, I'll update the description. I do think it seems easier to add here than to add a new function every time

plugin-runscanner/src/main/java/ca/on/oicr/gsi/shesmu/runscanner/RunScannerClient.java
grapigeau: No lanes for Ultima (well just one)

plugin-runscanner/src/main/java/ca/on/oicr/gsi/shesmu/runscanner/RunScannerClient.java
grapigeau: Nope, Nexus and MISO both omit the "-"

avarsava (approved):

plugin-runscanner/src/main/java/ca/on/oicr/gsi/shesmu/runscanner/RunScannerClient.java
```diff
@@ -229,6 +240,34 @@ public Optional<JsonNode> dragen_bclconvert(
...
+      UltimaWorkflowRun workflowRun = ultimaPipelineRun.get(workflow);
+      if (workflowRun != null) {
+        for (UltimaAnalysisUnit unit : workflowRun.getAnalysisOutputs()) {
+          if (unit.getBarcode().equals(barcode)) {
+            return Optional.ofNullable(MAPPER.valueToTree(unit.getFiles()));
+          }
+        }
+      }
+      return Optional.empty();
+    } catch (ClassCastException | InitialCachePopulationException e) {
```
callunity: catch JacksonException here too, for any deserialization errors? (To prevent them from escaping this function)

plugin-runscanner/src/main/java/ca/on/oicr/gsi/shesmu/runscanner/RunScannerClient.java
```diff
@@ -229,6 +240,34 @@ public Optional<JsonNode> dragen_bclconvert(
...
+      if (theRun.isEmpty()) return Optional.empty();
+      RunInformation runInfo = theRun.get();
+      if (runInfo.pipelineRun().isEmpty()) return Optional.empty();
+
+      UltimaPipelineRun ultimaPipelineRun = (UltimaPipelineRun) runInfo.pipelineRun().get();
+      UltimaWorkflowRun workflowRun = ultimaPipelineRun.get(workflow);
+      if (workflowRun != null) {
+        for (UltimaAnalysisUnit unit : workflowRun.getAnalysisOutputs()) {
+          if (unit.getBarcode().equals(barcode)) {
+            return Optional.ofNullable(MAPPER.valueToTree(unit.getFiles()));
```
callunity, line 262: I've found a bug in how run scanner's DTOs are currently constructed, such that this line will return DTOs with no `format` key. Here's the [PR in run scanner](https://github.com/miso-lims/runscanner/pull/233) which will include the `format` key in the JSON output of this function. It includes a method which you could use right now to get `format` in the JSON output, or we could plan to keep this PR as-is, merge the run scanner PR and release run scanner before the next Shesmu release, which would include this feature.

avarsava: Not really a bug, that field is only for jackson's sake. You have to request cram or emseq (metadata) when calling the function, so I think the format you'll get back is known already

grapigeau: It's a one to one mapping now ("CRAMGeneration", "cram") ("EmSeq", "metadata") but I'm not sure that will be the case going forward. 
I think getting the run scanner change in makes the most sense. Thanks for that Heather!

callunity (approved):

PR #1457, opened 2026-09-04
Fix pinery refresh failures

callunity:
JIRA Ticket: GP-5892

- [x] Updates Changelog

This is currently deployed to dev, and looks to have resolved the Pinery fetching issues.

### DIFF
--- a/changes/fix_cache_refresh_error_messages.md
+++ b/changes/fix_cache_refresh_error_messages.md
@@ -1,0 +1,3 @@
+Uninformative error messages when a cache failed to refresh
+* the whole exception cause chain is now reported instead of only the outermost message, which was frequently something as useless on its own as `closed`
+* a JSON list that stops part way through now reports how many records and how many bytes were read before it failed, which distinguishes a response that was cut short from malformed JSON
--- a/changes/fix_cache_refresh_error_messages.md
+++ b/changes/fix_cache_refresh_error_messages.md
@@ -1,3 +1,5 @@
 Uninformative error messages when a cache failed to refresh
 * the whole exception cause chain is now reported instead of only the outermost message, which was frequently something as useless on its own as `closed`
 * a JSON list that stops part way through now reports how many records and how many bytes were read before it failed, which distinguishes a response that was cut short from malformed JSON
+* each message in the chain is flattened onto one line and abbreviated, since Jackson's run to several lines and append a source location naming the features that were disabled while producing it
+* an `Error`, such as running out of memory, is no longer wrapped up as an ordinary failed refresh and swallowed by the cache machinery
--- a/changes/fix_pinery_eager_provenance_request.md
+++ b/changes/fix_pinery_eager_provenance_request.md
@@ -1,0 +1,3 @@
+Pinery requested sample provenance before it was ready to read it
+* both arguments to `Stream.concat` are evaluated before either is read, so the sample provenance request was issued while lane provenance was still being consumed, leaving a response body unread and its connection idle
+* the request is now deferred until the lane provenance has been read
--- a/plugin-pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
+++ b/plugin-pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
@@ -62,20 +62,25 @@ public class PinerySource extends JsonPluginFile<PineryConfiguration> {
       }
       final PineryConfiguration cfg = config.get();
       final Map<String, Integer> badSetCounts = new TreeMap<>();
-      final Map<String, RunDto> allRuns =
+      final Map<String, RunDto> allRuns;
+      // Collecting can fail on its own, on a run with no name, so the response body is closed
+      // explicitly rather than relying on reaching the end of the array to do it
+      try (final var runs =
           HTTP_CLIENT
               .send(
                   httpGet(cfg.getUrl() + "/sequencerruns", Optional.of(cfg.getTimeout())),
                   new JsonListBodyHandler<>(MAPPER, RunDto.class))
               .body()
-              .get()
-              .collect(
-                  Collectors.toMap(
-                      RunDto::getName,
-                      Function.identity(),
-                      // If duplicate run names occur, pick one at random, because the universe is
-                      // spiteful, so we spite it back.
-                      (a, b) -> a));
+              .get()) {
+        allRuns =
+            runs.collect(
+                Collectors.toMap(
+                    RunDto::getName,
+                    Function.identity(),
+                    // If duplicate run names occur, pick one at random, because the universe is
+                    // spiteful, so we spite it back.
+                    (a, b) -> a));
+      }
       final Set<Pair<String, String>> validLanes = new HashSet<>();
       // The samples must be read after the lanes, which populate validLanes, so the request for
       // them is deferred rather than left waiting on an idle connection
@@ -357,20 +362,25 @@ public class PinerySource extends JsonPluginFile<PineryConfiguration> {
       }
       final PineryConfiguration cfg = config.get();
       final Map<String, Integer> badSetCounts = new TreeMap<>();
-      final Map<String, RunDto> allRuns =
+      final Map<String, RunDto> allRuns;
+      // Collecting can fail on its own, on a run with no name, so the response body is closed
+      // explicitly rather than relying on reaching the end of the array to do it
+      try (final var runs =
           HTTP_CLIENT
               .send(
                   httpGet(cfg.getUrl() + "/sequencerruns", Optional.of(cfg.getTimeout())),
                   new JsonListBodyHandler<>(MAPPER, RunDto.class))
               .body()
-              .get()
-              .collect(
-                  Collectors.toMap(
-                      RunDto::getName,
-                      Function.identity(),
-                      // If duplicate run names occur, pick one at random, because the universe is
-                      // spiteful, so we spite it back.
-                      (a, b) -> a));
+              .get()) {
+        allRuns =
+            runs.collect(
+                Collectors.toMap(
+                    RunDto::getName,
+                    Function.identity(),
+                    // If duplicate run names occur, pick one at random, because the universe is
+                    // spiteful, so we spite it back.
+                    (a, b) -> a));
+      }
       final Set<Pair<String, String>> validLanes = new HashSet<>();
       // The samples must be read after the lanes, which populate validLanes, so the request for
       // them is deferred rather than left waiting on an idle connection
--- a/plugin-pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
+++ b/plugin-pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
@@ -22,6 +22,7 @@ import ca.on.oicr.gsi.status.SectionRenderer;
 import ca.on.oicr.ws.dto.*;
 import io.prometheus.client.Gauge;
 import java.io.IOException;
+import java.io.UncheckedIOException;
 import java.net.http.HttpClient;
 import java.nio.file.Path;
 import java.nio.file.Paths;
@@ -76,6 +77,8 @@ public class PinerySource extends JsonPluginFile<PineryConfiguration> {
                       // spiteful, so we spite it back.
                       (a, b) -> a));
       final Set<Pair<String, String>> validLanes = new HashSet<>();
+      // The samples must be read after the lanes, which populate validLanes, so the request for
+      // them is deferred rather than left waiting on an idle connection
       return Stream.concat(
               lanes(
                   cfg.getUrl(),
@@ -84,13 +87,15 @@ public class PinerySource extends JsonPluginFile<PineryConfiguration> {
                   badSetCounts,
                   allRuns,
                   (run, lane) -> validLanes.add(new Pair<>(run, lane))),
-              samples(
-                  cfg.getUrl(),
-                  cfg.getVersion(),
-                  cfg.getProvider(),
-                  badSetCounts,
-                  allRuns,
-                  (run, lane) -> validLanes.contains(new Pair<>(run, lane))))
+              lazily(
+                  () ->
+                      samples(
+                          cfg.getUrl(),
+                          cfg.getVersion(),
+                          cfg.getProvider(),
+                          badSetCounts,
+                          allRuns,
+                          (run, lane) -> validLanes.contains(new Pair<>(run, lane)))))
           .onClose(
               () ->
                   badSetCounts.forEach(
@@ -367,6 +372,8 @@ public class PinerySource extends JsonPluginFile<PineryConfiguration> {
                       // spiteful, so we spite it back.
                       (a, b) -> a));
       final Set<Pair<String, String>> validLanes = new HashSet<>();
+      // The samples must be read after the lanes, which populate validLanes, so the request for
+      // them is deferred rather than left waiting on an idle connection
       return Stream.concat(
               lanes(
                   cfg.getUrl(),
@@ -375,13 +382,15 @@ public class PinerySource extends JsonPluginFile<PineryConfiguration> {
                   badSetCounts,
                   allRuns,
                   (run, lane) -> validLanes.add(new Pair<>(run, lane))),
-              samples(
-                  cfg.getUrl(),
-                  cfg.getVersion(),
-                  cfg.getProvider(),
-                  badSetCounts,
-                  allRuns,
-                  (run, lane) -> validLanes.contains(new Pair<>(run, lane))))
+              lazily(
+                  () ->
+                      samples(
+                          cfg.getUrl(),
+                          cfg.getVersion(),
+                          cfg.getProvider(),
+                          badSetCounts,
+                          allRuns,
+                          (run, lane) -> validLanes.contains(new Pair<>(run, lane)))))
           .onClose(
               () ->
                   badSetCounts.forEach(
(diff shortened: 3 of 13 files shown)


changes/fix_cache_refresh_error_messages.md
```diff
@@ -0,0 +1,5 @@
+Uninformative error messages when a cache failed to refresh
+* the whole exception cause chain is now reported instead of only the outermost message, which was frequently something as useless on its own as `closed`
+* a JSON list that stops part way through now reports how many records and how many bytes were read before it failed, which distinguishes a response that was cut short from malformed JSON
```
avarsava: This sounds fine as a feature, but unless we can know how many bytes are expected, does it really distinguish anything?

changes/fix_cache_refresh_error_messages.md
```diff
@@ -0,0 +1,5 @@
...
+* each message in the chain is flattened onto one line and abbreviated, since Jackson's run to several lines and append a source location naming the features that were disabled while producing it
+* an `Error`, such as running out of memory, is no longer wrapped up as an ordinary failed refresh and swallowed by the cache machinery
```
avarsava: What is it instead?

changes/fix_cache_refresh_error_messages.md
```diff
@@ -0,0 +1,5 @@
+Uninformative error messages when a cache failed to refresh
+* the whole exception cause chain is now reported instead of only the outermost message, which was frequently something as useless on its own as `closed`
+* a JSON list that stops part way through now reports how many records and how many bytes were read before it failed, which distinguishes a response that was cut short from malformed JSON
+* each message in the chain is flattened onto one line and abbreviated, since Jackson's run to several lines and append a source location naming the features that were disabled while producing it
```
avarsava: Disagree with doing this. We shouldn't arbitrarily cut off error messages because one library's run long.

plugin-pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
```diff
@@ -62,21 +63,28 @@ protected Stream<PineryIUSForAnalysisValue> fetch(Instant lastUpdated) throws Ex
       }
       final PineryConfiguration cfg = config.get();
       final Map<String, Integer> badSetCounts = new TreeMap<>();
-      final Map<String, RunDto> allRuns =
+      final Map<String, RunDto> allRuns;
+      // Collecting can fail on its own, on a run with no name, so the response body is closed
+      // explicitly rather than relying on reaching the end of the array to do it
+      try (final var runs =
```
avarsava: Please don't use `var`

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/json/JsonListBodyHandler.java
```diff
@@ -24,56 +27,134 @@ private static <W> HttpResponse.BodySubscriber<Supplier<Stream<W>>> asJSON(
...
+    try {
+      return parser.currentLocation().getByteOffset();
+    } catch (RuntimeException e) {
+      return -1;
+    }
+  }
+
   private static <W> Supplier<Stream<W>> toSupplierOfType(
       JsonMapper jsonMapper, InputStream inputStream, JavaType targetType) {
+    return () -> streamOf(jsonMapper, inputStream, targetType);
```
avarsava: Why make this two methods?

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/json/JsonListBodyHandler.java
```diff
@@ -24,56 +27,134 @@ private static <W> HttpResponse.BodySubscriber<Supplier<Stream<W>>> asJSON(
...
+                    final W item;
+                    try {
+                      if (parser.nextToken() == JsonToken.END_ARRAY) {
+                        parser.close();
+                        return false;
+                      }
+                      item = itemReader.readValue(parser);
+                    } catch (RuntimeException e) {
+                      // Only RuntimeException; wrapping an Error would turn it into something the
+                      // cache machinery catches and reports as a mere failed refresh. An Error
```
avarsava: Wouldn't we want it to be treated as a failed refresh?

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/json/JsonListBodyHandler.java
```diff
@@ -24,56 +27,134 @@ private static <W> HttpResponse.BodySubscriber<Supplier<Stream<W>>> asJSON(
...
+        // if the server had sent broken JSON. Only a parse failure is treated this way; a body
+        // that dies mid-read throws JacksonIOException and is left to report itself as such.
+        throw new IllegalArgumentException(
+            String.format(
+                "Expected a JSON array of %s but the response was not JSON",
+                targetType.getRawClass().getSimpleName()),
+            e);
+      }
+      if (firstToken == JsonToken.START_ARRAY) {
+        releaseParser = false;
```
avarsava: why would we not release the parser in this case?

changes/fix_pinery_eager_provenance_request.md
```diff
@@ -0,0 +1,3 @@
+Pinery requested sample provenance before it was ready to read it
+* both arguments to `Stream.concat` are evaluated before either is read, so the sample provenance request was issued while lane provenance was still being consumed, leaving a response body unread and its connection idle
```
avarsava: Is this true? The Javadoc for Stream says concat "Creates a lazily concatenated stream": https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/stream/Stream.html#concat(java.util.stream.Stream,java.util.stream.Stream)

avarsava (commented):
That's a lot of changelog. Could there be less changelog?

changes/fix_cache_refresh_error_messages.md
callunity: It allows us to compare how many bytes were received across different requests - a consistently repeated number of bytes would suggest a fixed-sized buffer somewhere (Pinery), and different numbers would point towards a timeout.

changes/fix_cache_refresh_error_messages.md
callunity: It's raised as an exception: https://github.com/oicr-gsi/shesmu/pull/1457/changes#diff-e07ede96c9da462e363f566cc77bcb39f3ff46a961e2e7f526ed05401834e867R116-R128

avarsava: So the difference is that the stack trace prints? It still gets rescheduled, right?

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/json/JsonListBodyHandler.java
callunity: because if it's a START_ARRAY token, the method returns a stream which has methods for reading the inner items before releasing the parser. (Releasing the parser releases the HTTP connection shouldn't happen right at the start of the JSON array - see comment on 68-70).

changes/fix_cache_refresh_error_messages.md
callunity: Yes, the next refresh does do the fetch again.

changes/fix_pinery_eager_provenance_request.md
callunity: The arguments (`lanes()` and `samples()`) get evaluated eagerly into their constituent streams before `concat` gets called, at which point the two streams are lazily concatenated. I'm not sure if this changelog is actually helpful.

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/json/JsonListBodyHandler.java
callunity: Looks like Claude chose this for readability (less indentation), for making the `streamOf` method package-private for better testing (but could get this by changing `toSupplierOfType` to package-private), and because moving the Javadoc currently on `streamOf` to `toSupplierOfType` would have it describing the return value of the return value. I don't find these to be especially compelling arguments, and can move it all back under `toSupplierOfType` if you'd prefer.

plugin-pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
callunity: updated

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/json/JsonListBodyHandler.java
avarsava: Sure, personally I find the multiple jumps harder to follow

changes/fix_cache_refresh_error_messages.md
callunity: The abbreviation is 200 characters per error message. With this change, the current refresh error messages would be rendered as:
```
Exception occurred while refreshing cache pinery /path/to/miso.pinery JsonListReadException: Failed reading JSON array of LaneProvenanceDto after 47231 record(s) and 20971520 byte(s) caused by JacksonIOException: closed at [No location information] (through reference chain: ca.on.oicr.ws.dto.LaneProvenanceDto["sequencerRunAttributes"]) caused by IOException: closed caused by IOException: chunked transfer encoding, state: READING_DATA caused by EOFException: EOF reached while reading is as follows:
    <STACKTRACE>
```
The full stacktrace is still there, containing any long lines.

shesmu-pluginapi/src/main/java/ca/on/oicr/gsi/shesmu/plugin/json/JsonListBodyHandler.java
callunity: updated

changes/fix_cache_refresh_error_messages.md
avarsava: I still don't see the benefit of doing this, while the risk is dropping important information from the error message, which may contain important context for an error

plugin-pinery/src/test/java/ca/on/oicr/gsi/shesmu/pinery/LazilyTest.java
```diff
@@ -0,0 +1,155 @@
...
+  public void testRequestIsDeferredUntilRead() {
+    final List<String> events = new ArrayList<>();
+    final Stream<String> combined =
+        Stream.concat(
+            Stream.of("a1", "a2").peek(events::add),
+            PinerySource.lazily(
+                () -> {
+                  events.add("requested");
+                  return Stream.of("b1").peek(events::add);
+                }));
```
avarsava, line 32: `
Stream.concat(
  Stream.of("a1", "a2").peek(events::add),
  Stream.of("requested", "b1").peek(events.add))
`
Also passes the test, so I'm not sure what this is proving?

changes/fix_cache_refresh_error_messages.md
callunity: I've removed truncation.

plugin-pinery/src/test/java/ca/on/oicr/gsi/shesmu/pinery/LazilyTest.java
callunity: Here's the behaviour that would occur without `PinerySource.lazily()`:
```
public Stream<String> nonLazyFunction(List<String> events) {
    events.add("requested");
    return Stream.of("b1").peek(events::add);
  }

  /**
   * A non-lazy argument will be evaluated before the preceding batch is read, which is expensive
   * when the non-lazy argument contains an HTTP call,
   */
  @Test
  public void testNonLazyRequestNotDeferred() {
    final List<String> events = new ArrayList<>();
    final Stream<String> combined =
        Stream.concat(Stream.of("a1", "a2").peek(events::add), nonLazyFunction(events));
    assertEquals(
        List.of("requested"), events, "non-lazy function gets run before the stream is read");
    combined.forEach(item -> {});
    assertEquals(List.of("requested", "a1", "a2", "b1"), events);
  }
```
Recall that we started with `Stream.concat(lanes(), samples())`. The `lazily` function pushes the evaluation of `samples()` (includes its HTTP call) until after `lanes()` has been evaluated and its stream closed. Without it, `samples()` got called while `lanes()` was still being consumed, leaving the `samples()` connection open for a while before it got read and running the risk of request timeout.

plugin-pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
```diff
@@ -740,6 +759,36 @@ private static boolean isRunValid(RunDto run) {
     return run != null && run.getCreatedDate() != null;
   }
 
+  /** A batch of provenance that has to be requested over HTTP before it can be read */
+  interface ProvenanceSource<T> {
```
avarsava, line 763: Isn't this a reimplementation of Supplier? https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/function/Supplier.html

plugin-pinery/src/test/java/ca/on/oicr/gsi/shesmu/pinery/LazilyTest.java
avarsava: OK, I see. It's not so much that arguments to Stream.concat are always eagerly evaluated so much as the JVM necessarily has to execute any elements in the method call that aren't Stream-building, such as the List add call. The Stream in the return line is not evaluated, or b1 would be in events. And that HTTP fetching activity is indeed not part of the Stream library, so that tracks.

Seems like wrapping in Stream.of and flatmapping after _is_ the simplest way to change samples() and lanes() to look like pure Stream-building, but I wonder if changing samples() to, for instance:

```
return Stream.of(HTTP_CLIENT
          .send(
              httpGet(
                  baseUrl + "/provenance/v" + version + "/sample-provenance",
                  config.map(PineryConfiguration::getTimeout)),
              new JsonListBodyHandler<>(MAPPER, SampleProvenanceDto.class))
          .body()
          .get())
          .flatMap((a) -> a)
          .filter(
```

(and a similar change for lanes()) would accomplish the same thing with less complexity?

If not, I would at least like the inline comments clarified.

callunity: `Stream.of` + `flatMap` isn't sufficient to defer the HTTP call:
```
public Stream<String> fakeSendStream(List<String> events) {
    return Stream.of(
            fakeSend(events)
                .body()
                .get())
        .flatMap(a -> a);
  }

  // Stand-in for HTTP_CLIENT.send(...) returning something HttpResponse-like
  private FakeHttpResponse<Stream<String>> fakeSend(List<String> events) {
    events.add("requested");
    return new FakeHttpResponse<>(Stream.of("b1").peek(events::add));
  }

  private static class FakeHttpResponse<T> {
    private final FakeBody<T> body;
    FakeHttpResponse(T value) { this.body = new FakeBody<>(value); }
    FakeBody<T> body() { return body; }
  }

  private static class FakeBody<T> {
    private final T value;
    FakeBody(T value) { this.value = value; }
    T get() { return value; }
  }

  /**
   * Wrapping the non-lazy argument in {@code Stream.of(...).flatMap(a -> a)} does not defer it. 
   * Deferring needs a lambda, which is what
   * {@link PinerySource#lazily(PinerySource.ProvenanceSource)} takes.
   */
  @Test
  public void testFlatMappedRequestNotDeferred() {
    final List<String> events = new ArrayList<>();
    final Stream<String> combined =
        Stream.concat(
            Stream.of("a1", "a2").peek(events::add),
            fakeSendStream(events));
    assertEquals(
        List.of("requested"),
        events,
        "wrapping in Stream.of(...).flatMap(a -> a) does not defer the request");
    combined.forEach(item -> {});
    assertEquals(List.of("requested", "a1", "a2", "b1"), events);
  }
```

plugin-pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
callunity: Yes, but with the `throws` clause, which Supplier's `get` does not have. The `samples()` call throws IOException, InterruptedException. Using bare `Supplier` means wrapping the calls to `samples()` in try/catch, but using this custom interface means `lazily` handles the exceptions in one place, rather than wrapping each call to `samples` in try/catch + error handling. 

What do you think about renaming the interface for clarity, to something like `ThrowingSupplier` ?

plugin-pinery/src/test/java/ca/on/oicr/gsi/shesmu/pinery/LazilyTest.java
avarsava: OK. It doesn't have to be literally that. I mean is there some way to inline the behaviour into samples() and lanes()?

> If not, I would at least like the inline comments clarified.

plugin-pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
avarsava: If there's no avoiding having this interface, then yes please

plugin-pinery/src/test/java/ca/on/oicr/gsi/shesmu/pinery/LazilyTest.java
callunity: Only the second arg to `Stream.concat` (`samples()`, at this time) needs to be deferred until after the first arg's stream is consumed. If we inline the behaviour into `samples()` then it looks just like `lanes()` at the call sites, and makes it look like there's something inherently different about the `samples()` call that requires it, when it's only because of its positioning within the args to `Stream.concat` that has `samples()` being wrapped in `lazily()` at this time. 

I sorry, but I don't understand which inline comments you're referring to as requiring clarification...

plugin-pinery/src/main/java/ca/on/oicr/gsi/shesmu/pinery/PinerySource.java
callunity: As I said it is possible to avoid having this interface, but I think that keeping the exception handling in one place justifies its usage. I'll rename it.

avarsava (approved):

mlaszloffy (approved):

grapigeau (approved):

PR #1463, opened 2026-09-17
Fix issue where timestamps from run scanner were deserialized

callunity:
JIRA Ticket: GP-5047

- [x] Updates Changelog


mlaszloffy (approved):

avarsava:
What change should I be seeing? I deployed this change and do not see any change in the dragen_bclconvert function's output.

Using epochconverter.com, the example value `1775129153.0499983` in your commit message reasonably parses as April 2 2026 if the converter is "Assuming that this timestamp is in **seconds**", but if you switch that to nanoseconds, it's beyond the current epoch time. So it seems ok?

callunity:
> What change should I be seeing? I deployed this change and do not see any change in the dragen_bclconvert function's output.
> 
> Using epochconverter.com, the example value `1775129153.0499983` in your commit message reasonably parses as April 2 2026 if the converter is "Assuming that this timestamp is in **seconds**", but if you switch that to nanoseconds, it's beyond the current epoch time. So it seems ok?

I have fixed the output now - I'm really sorry about that. 

It's possible to convert the example value, and you've shown that Vidarr can, but I think that there's some benefit to consistent representation of timestamps across the project.

avarsava:
> It's possible to convert the example value, and you've shown that Vidarr can, but I think that there's some benefit to consistent representation of timestamps across the project.

My confusion was more that your change disables writing timestamps as nanoseconds, but the value was seconds, not nanoseconds.

Anyway, it's in milliseconds now. Change looks ok, but I will need to make sure nothing on the import path wants specifically seconds.

avarsava (approved):

callunity:
> > It's possible to convert the example value, and you've shown that Vidarr can, but I think that there's some benefit to consistent representation of timestamps across the project.
> 
> My confusion was more that your change disables writing timestamps as nanoseconds, but the value was seconds, not nanoseconds.
> 
> Anyway, it's in milliseconds now. Change looks ok, but I will need to make sure nothing on the import path wants specifically seconds.

The value was in seconds because that's how Jackson represents a nanosecond-precision timestamp (as a long would overflow), but milliseconds can be represented in an integer.

